PR #11, opened 2015-10-22
Introduces BasiGX.util.Map

KaiVolland:


src/util/Map.js
```diff
@@ -0,0 +1,40 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+/**
+ *
+ * Layer Util
```
marcjansen: API docs are not correct

src/util/Map.js
```diff
@@ -0,0 +1,40 @@
...
+ *
+ */
+Ext.define('BasiGX.util.Map', {
+
+    statics: {
+        getResolution: function (map) {
+            return map.getView().getResolution();
+        },
+
+        getScale: function(map){
```
marcjansen: Do you want to mention the source of this?

marcjansen:
Two minor comments which you may want to address. Good to go otherwise. Thanks.

PR #19, opened 2015-11-05
added tests for wfscluster, animate, application and configparser

weskamm:
I added some more tests for the BasiGX package. Configparser is still rough but should be ok since it will get reworked soon i believe


KaiVolland:
Tests can't be bad. :+1: :hocho: :dancers:

marcjansen:
Should this be rebased? The changes to `src/plugin/WfsCluster.js` are already in master, right?

marcjansen:
Looks good to me. Please merge @weskamm

PR #20, opened 2015-11-17
made template in Layersetchooser configurable

carmen123:
title says it all


weskamm:
this one lgtm, more config options are always welcome

PR #22, opened 2015-11-22
Added tagfinder to overpass search; several optimizations

weskamm:
Added the tagfinder API to the search, effectively reducing the search waiting time due to more concrete queries against overpass API. The user now gets a list of available tags with description for his searchterms, Overpass query gets executed with the selected offical osm tag after the user has selected one.


marcjansen:
Looks good to me, @weskamm.

Note to self (and others): we need examples, tests and i18n consistency. I'll open tickets.

marcjansen:
See #23, #24 and #25.

weskamm:
Thanks for review and reminder! Will merge...

PR #28, opened 2015-11-23
Fixed problem on detecting wms-layer-styles in capabilities

weskamm:
This fixes breaking of the component when capabilities return a layer without style information


marcjansen:
Yes, please. Please merge.

PR #29, opened 2015-11-23
Made the buttons usable for modern client

weskamm:
This PR changes all current buttons to extend from `Ext.Button` instead of `Ext.button.Button`.
This way they are compatible with the classic and modern toolkit.
There are still buttons which cannot be used in the modern toolkit (Permalink, AddWms and CoordinateTransform), because they rely on classic components like windows.

Other changes involved are the usage of the constructor and moving the button handlers to the config object. 
The tooltip binding had to be taken special care of because the "modern" buttons dont support the `setToolTip` method.

### DIFF
--- a/src/view/button/AddWms.js
+++ b/src/view/button/AddWms.js
@@ -22,7 +22,7 @@
  * @class BasiGX.view.button.AddWms
  */
 Ext.define("BasiGX.view.button.AddWms", {
-    extend: "Ext.button.Button",
+    extend: "Ext.Button",
     xtype: 'basigx-button-addwms',
 
     requires: [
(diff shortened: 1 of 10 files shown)


marcjansen:
Just to get this right: they now work on **both** modern and classic?

weskamm:
Yes. Another PR will follow the next days so that the icons in modern toolkit are also visible, currently they are not because of no support for glyphs in modern toolkit

marcjansen:
Pnce I have time to cjeck this in my apps I'll :+1: this. If it is urgent, please ping e.g. @KaiVolland

weskamm:
Take your time :+1:

weskamm:
Display of icons has been fixed. Modern toolkit uses html, classic uses glyph.
Tested in modern and classic app, works as intended.

marcjansen:
Very nice work, applied cleanly and works fine in two other apps. Thanks, @weskamm, please merge.

weskamm:
thx, will merge

PR #30, opened 2015-11-25
Added a window-like panel for modern applications

weskamm:
This panel can be used to imitate a window like component. It has a header which can have a title and also serves a closable "X" button to close the panel


KaiVolland:
love it! :dancers:

Issue #25, opened 2015-11-23
Make hardcoded UI-strings translatable


marcjansen:
Currently a lot of string that'll end up visible to the user are not configurable or translatable. We need to change that.

weskamm:
I had to add requires for the inline viewmodels of our classes in #32 to get the modern builds working.
These viewmodels, binds and requires should be removed when implementing a translation functionality in the future.

dnlkoch:
See #68.

PR #32, opened 2015-11-30
Added missing viewModel requires

weskamm:
Requires of `Ext.app.ViewModel` have been added to enable production builds on modern toolkits


marcjansen:
:+1: very good that you discuss the removal of this dependency in #25.

Please merge.

weskamm:
thx!

PR #33, opened 2015-12-09
Sort apps, formats and layouts alphabetically

marcjansen:
This PR suggest that the print-combos for `apps`, `formats` and `layouts` are sorted alphabetically in ascending order.

I decided against configurabillity, since I cannot imagine a case where we want to have descending order.

The methods used to sort are diffenrent, as `apps` and `formats` are plain arrays, while `layouts()` actually gives us a store. The store can be soreted against any field of `GeoExt.data.model.print.Layout`, useful in our case s the name.

Please review and merge at your own will.


weskamm:
+1, will merge

marcjansen:
Thanks, @weskamm

PR #34, opened 2015-12-09
Fire events while adding attributefields

marcjansen:
This PR suggests adding the events `beforeattributefieldsadd` and `attributefieldsadd`, so that subclasses have a chance to alter the generated form, e.g. by hiding fields (or not adding them at all) or setting values.

Please review and merge at will.


weskamm:
Nice addition and very welcome, especially the before handler. Merging..

marcjansen:
Thanks again ☺

PR #36, opened 2015-12-14
Read layer properties for hovering from fields

marcjansen:
This makes it easier for applications to determine the fields
which are used to controll hovering. If, for example you want
to disable hovering of a layer temporarily, you can now just
determine the correct property, and then set the value for it
to `false`.

This commit also adds basic tests that document the behaviour.

Once this is merged, we cannot 100% guarantee, that `hoverfield`
and `hoverable` properties exist on our layer instances.
Applications should now use the static fields from the
hoverplugin.

If an application has the need for different fields for hovering
the plugin class can be extended and the statics can be reassigned
to match the new environment. See the tests for an example of
this.

### DIFF
--- a/src/plugin/Hover.js
+++ b/src/plugin/Hover.js
@@ -24,7 +24,23 @@ Ext.define('BasiGX.plugin.Hover', {
 
     inheritableStatics: {
         HOVER_OVERLAY_IDENTIFIER_KEY: 'name',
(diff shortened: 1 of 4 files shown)


weskamm:
lgtm, please merge!

marcjansen:
Nice and thanks!

PR #37, opened 2015-12-17
Make source & behaviour of selection configurable

marcjansen:
This introduces two new configuration options for the hover plugin:

* `selectMulti`, and
* `selectEventOrigin`

With `selectMulti` you can control whether the select interaction may ever
possibly select more than one feature from the hover layer. The default is
`true` for backwards compatibillity.

With `selectEventOrigin` you can control if you want the origin of the
click event firing to be the collection of the select interaction (`add`
event, also the default for historical reasons) or the interaction itself
(its `select` event).

This PR shouldn't introduce any regressions, as the configs are fully optional.

Please review.

### DIFF
--- a/src/plugin/Hover.js
+++ b/src/plugin/Hover.js
@@ -53,12 +53,39 @@ Ext.define('BasiGX.plugin.Hover', {
         hoverVectorLayerInteraction: null
     },
 
+    /**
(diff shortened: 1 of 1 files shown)


KaiVolland:
:+1: :dancers: Nice work. Plz merge.

marcjansen:
Cheers for the review.

PR #44, opened 2016-01-15
Added CSRF header util

weskamm:
Will get more and more usable in the future when SHOGun2 gets spread..

### DIFF
--- a/src/util/CSRF.js
+++ b/src/util/CSRF.js
@@ -1,0 +1,32 @@
+Ext.define('BasiGX.util.CSRF', {
+
+    statics: {
+        /**
+        * Get the CSRF token value.
+        */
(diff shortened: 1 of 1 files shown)


buehner:
:+1:

marcjansen:
A nice addition. Some minor remarks:

* The names shoud be configurable and default to the SHOGun2 values.
* The class doesn't have the GPL-header
* and the returns aren't documented.

Please merge once these are adressed.

marcjansen:
You are too fast, guys.

weskamm:
thanks @marcjansen, PR #45 adresses the mentioned issues.
The names you are talking about are the defaults given by spring security, so i kept them as is.

PR #46, opened 2016-01-22
Change default of LAYER_HOVERFIELD_PROPERTY_NAME

weskamm:
CamelCase usage of this property is the better choice as its already used in several backend-models in our applications. I tried an override first as we often extend the Hover class, but that leads to a lot of code and problems because of the early stage when the configparser handles this field. So maybe its easier to just change this property.

@marcjansen: Please check if this gets you in trouble :-)

### DIFF
--- a/src/plugin/Hover.js
+++ b/src/plugin/Hover.js
@@ -40,7 +40,7 @@ Ext.define('BasiGX.plugin.Hover', {
          *
          * @type {String}
          */
-        LAYER_HOVERFIELD_PROPERTY_NAME: 'hoverfield'
+        LAYER_HOVERFIELD_PROPERTY_NAME: 'hoverField'
     },
 
     config: {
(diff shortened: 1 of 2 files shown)


marcjansen:
I need to double check on monday. If this blocks your work, feel free to merge.

On a second thought: why is this even a `static` and not simply a `config`? This would make overriding / configuring very easy and we still could have a default.

weskamm:
| On a second thought: why is this even a static and not simply a config

See:
https://github.com/weskamm/BasiGX/commit/f5f442ae7e096463fe0619d7fe1fc985d4fdad3f

:-)

marcjansen:
I checked this and and this doesn't break my precious applications. Go ahead and merge, @weskamm !

marcjansen:
Mr fast-lane!

weskamm:
^^

PR #47, opened 2016-02-03
Fixed coordinatetransform for ol-production builds

weskamm:
Changed the detection of coordinates in events as `ol.MapBrowserPointerEvent` is not available in production builds of ol3


marcjansen:
Please rebase!

marcjansen:
:black_joker:

PR #48, opened 2016-02-03
Fix the tests by using polyfill for phantomjs

weskamm:
Current tests fail on master of ol3 as the phantomjs aint capable of using the `requestAnimationFrame` function. This PR fixes this by adding an appropriate polyfill for phantomjs


marcjansen:
**Wicked!** :+1: will merge once travis is green.

PR #49, opened 2016-02-10
Added Object util

weskamm:
Added new util to work with Objects. Currently supports getting a specific value from a nested Object


KaiVolland:
NICE!



Please merge!

PR #50, opened 2016-02-15
Added fallback for IE9 in treeColor

weskamm:
Fixes a minor coloring issue in IE9

### DIFF
--- a/src/view/panel/LegendTree.js
+++ b/src/view/panel/LegendTree.js
@@ -202,6 +202,7 @@ Ext.define("BasiGX.view.panel.LegendTree", {
                 '.' + elemenIdAndCssClass + ' { ' +
                 ' background-color: ' + color + ';' +
                 ' background:linear-gradient(to right, white, ' + color + ');' +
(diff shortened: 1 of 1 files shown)


src/view/panel/LegendTree.js
```diff
@@ -202,6 +202,7 @@ Ext.define("BasiGX.view.panel.LegendTree", {
                 '.' + elemenIdAndCssClass + ' { ' +
                 ' background-color: ' + color + ';' +
                 ' background:linear-gradient(to right, white, ' + color + ');' +
+                ' filter: alpha(opacity=50)\10;' +//IE9 fallback
```
marcjansen: What does `\10` do?

marcjansen:
I don't understand the syntax used, @weskamm. Can you answer my question?

weskamm:
Its based on this advice and has been tested in IE9 - works:

http://stackoverflow.com/questions/11829792/ie-9-support-both-filter-and-rgba

marcjansen:
Thanks, merge at will.

PR #51, opened 2016-02-18
Removed the incorrect mapping from startbbox to restricted extent

weskamm:
Like the title says, i removed the invalid mapping from startbbox to (restricted) extent.
Use center and zoom to set your maps initial start appearance


marcjansen:
Good to go. Thanks @weskamm.

PR #52, opened 2016-02-18
Fixed layerorder while parsing shogun context

weskamm:
This fixes the ordering of layers that are contained in folders when using SHOGun 1 as backend serving the application context


marcjansen:
:bowtie: Please merge.

weskamm:
Thanks, merging!

PR #53, opened 2016-02-22
Use attributes default values from print-servlet

KaiVolland:
MapFish-Print 3.4. supports default values for attributes
so we use it.


weskamm:
nice 1! will merge..

PR #54, opened 2016-02-23
Minor query bugfix.

KaiVolland:


marcjansen:
These can already be configured by specifying a `legendLayerFilter` in an application override.

KaiVolland:
You are right. I didn't thought on that one. And then didn't thought i can use it in my project. Now i do. So i removed the stuff you mentioned and transformed this to a tiny PR. :)

marcjansen:
Now that is cool! Thanks for reconsidering.

marcjansen:
Please merge

PR #56, opened 2016-02-29
Adds one more "if" to ensure correct call of "renderPrintExtent."

KaiVolland:


weskamm:
More if, more safety, :+1:

PR #57, opened 2016-03-01
Fix autodeployment of BasiGX

KaiVolland:
This skips the build of jsduck cause it crashes the and forfeit the merge into the jsduck branch.


weskamm:
works for me if this fixes the build issue :+1:

PR #58, opened 2016-03-03
rewrite of movefeature method

weskamm:
This PR refactors the moveFeature method. It is now able to handle different geometry types, has been tested with points, linestrings and polygons. We do now execute a `doneFunction` when the animation has finished.
Pixels are used for the feature displacement, the translate respects the current views resolution


marcjansen:
:+1:  please merge.

PR #59, opened 2016-03-03
Introducing a msgbox util

weskamm:
Title says it all


marcjansen:
:+1: 

You know that I'd love to have tests here, but at a low coverage as we have right niw this would really be nit-pick. Please merge.

But at some point we need to start testing the cod here and elsewhere!

PR #60, opened 2016-03-08
Folders visible and expanded state

weskamm:
This PR fixes an issue with SHOGun 1 appContext: 
Folders that contain checked / visible layers will now also get checked and set to visible. The expanded state of a node needs to be propagated to get used in the treepanel again to correctly show the state of the gizmo treepanel.

### DIFF
--- a/src/util/ConfigParser.js
+++ b/src/util/ConfigParser.js
@@ -746,6 +746,7 @@ Ext.define('BasiGX.util.ConfigParser', {
                 } else if (node.leaf === false) {
                     // handling folders
                     var folder = me.createFolder(node);
+                    folder.set('expanded', node.expanded);
 
                     if (parent) {
                         parent.getLayers().push(folder);
--- a/src/util/ConfigParser.js
+++ b/src/util/ConfigParser.js
@@ -749,6 +749,7 @@ Ext.define('BasiGX.util.ConfigParser', {
 
                     if (parent) {
                         parent.getLayers().push(folder);
+                        folder.set('parentFolder', parent);
                     } else {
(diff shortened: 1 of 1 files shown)


src/util/ConfigParser.js
```diff
@@ -746,9 +746,11 @@ Ext.define('BasiGX.util.ConfigParser', {
                 } else if (node.leaf === false) {
                     // handling folders
                     var folder = me.createFolder(node);
+                    folder.set('expanded', node.expanded);
 
                     if (parent) {
                         parent.getLayers().push(folder);
+                        folder.set('parentFolder', parent);
```
marcjansen: Can you use `GeoExt.util.Layer.findParentGroup` instead?

src/util/ConfigParser.js
```diff
@@ -770,6 +772,14 @@ Ext.define('BasiGX.util.ConfigParser', {
...
                     if (parent) {
                         parent.getLayers().getArray().unshift(me.createLayer(mergedNode));
+                        if (node.checked) {
+                            parent.set('visible', true);
+                            var nextParent = parent.get('parentFolder');
+                            while (nextParent) {
+                                nextParent.set('visible', true);
+                                nextParent = nextParent.get('parentFolder');
+                            }
+                        }
```
marcjansen: Please check if the `folderToggleMode` may be of help

weskamm:
No, as this happens in the configparser and its to early to use util methods that rely on map or ol.layer.base

marcjansen:
OK then :+1: please merge.

PR #61, opened 2016-03-10
reenable cross origin

carmen123:
This PR reenables the crossOrigin attribute on ol.layer.sources.
This has been commented out for no reason, now it gets set dependent on the layer config.


marcjansen:
Thanks!

PR #62, opened 2016-03-11
added tilegrid to tiled layers

weskamm:
PR adds an ol.tilegrid to tiled layers, based on the maps extent and the given resolutions from appcontext. This is needed to avoid some kind of "shrinking" behaviour of tiles on certain zoom levels.


KaiVolland:
Cool stuff. Please merge.

PR #67, opened 2016-03-18
Be more cautious when creating a tilegrid

marcjansen:
Since 71e66d5ee6ef7bd70f94538f998ff3df3fb3eeba we overwrite a layers `tileGrid` configuration without checking if …

1. … such a `tileGrid` was already defined in the configuration
2. … the needed inputs for the generation are defined in the `mapConfig`

For 1. we should respect what we were given, and for 2. we should check if the key `resolutions` exists in `appContext.mapConfig`.

Additionally we should really be logging that something unexpected happened in the method `convertStringToNumericArray` instead of silently returning the input.

This is a follow-up for PR #62.

### DIFF
--- a/src/util/ConfigParser.js
+++ b/src/util/ConfigParser.js
@@ -549,8 +549,11 @@ Ext.define('BasiGX.util.ConfigParser', {
                 };
 
                 // set a tilegrid if needed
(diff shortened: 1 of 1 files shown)


marcjansen:
I will merge this without formal review as I consider this an urgent hotfix. Revert if you disagree.

PR #68, opened 2016-04-04
Language switcher

dnlkoch:
This PR suggests to move all language strings used in the ExtJS views towards inline viewModels and to bind the values of the model entries to the related class configs (where possible).

Havin' this realized, one can easily switch the language of an entire application using the `BasiGX.view.combo.Language` component (see example in this PR).


src/view/button/CoordinateTransform.js
```diff
@@ -89,10 +91,5 @@ Ext.define("BasiGX.view.button.CoordinateTransform", {
      */
     constructor: function(config) {
         this.callParent([config]);
-        if (this.setTooltip) {
```
weskamm: Have you checked the removal of this code against a mobile build with BasiGX? I have added theses checks because of incompatibilities with the modern toolkit. I can check the behaviour for you in a mobile application, just give me a sign

weskamm:
Lots of changes here, good work! I want to test this in several applications (modern/classic) before we can merge this. I will give a follow up comment when i have done that

marcjansen:
Looks good to me once @weskamm has confirmed that the `setTooltip`-thingy doesn't break mobile applications.

Thanks a ton for this!

weskamm:
looking all good to me, :+1: , plz merge

marcjansen:
Ok, let's get this in now.

Thanks everyone!

PR #70, opened 2016-04-06
Hiding the redline vector layer in the layertree

weskamm:
Usually, the redline vector layer should be hidden, so we set the specific property on instanciaton.
The treestore still needs the filter respecting this property for the filtering to work


KaiVolland:
Looking good. Please merge.

PR #74, opened 2016-04-14
Printconfig

KaiVolland:
- Introduces the missing functions for NorthArrow and ScaleBar.
- Allow printattribute stringfields to be empty.

### DIFF
--- a/src/view/form/Print.js
+++ b/src/view/form/Print.js
@@ -696,6 +696,7 @@ Ext.define("BasiGX.view.form.Print", {
      */
     getNorthArrowObject: function() {
         var northArrowObject = {};
+        // This file is located right beneath the config.yaml
         northArrowObject.graphic = "file://NorthArrow_10.svg";
         northArrowObject.backgroundColor = "rgba(0, 0, 0, 0)";
         return northArrowObject;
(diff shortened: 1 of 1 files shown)


marcjansen:
Thanks. Please merge when green.

src/view/form/Print.js
```diff
@@ -691,17 +691,29 @@ Ext.define("BasiGX.view.form.Print", {
...
     },
 
+    /**
+     * Creates a NorthArrow-Object
+     */
     getNorthArrowObject: function() {
-        // TODO
         var northArrowObject = {};
-
+        northArrowObject.graphic = "file://NorthArrow_10.svg";
```
weskamm: How does this work? Maybe add a comment line that the file has to be placed at specific location

marcjansen: I always thouight thias is a file that comes with MapFish Print v3?

weskamm: I dont't know so i am asking. I just want to avoid problems with broken resources. If this is the way how the servlet works, :+1: from me

KaiVolland:
I added a small comment to clearify the `"file://NorthArrow_10.svg"; `.

The configuration in this PR is more or less a workaround as me and @dnlkoch didn't discover any different solution to enable disable NorthArrow and ScaleBar.

Currently we set every used color for each - NorthArrow and ScaleBar - to transparent, and overwrite it from our config send from BasiGX.

weskamm:
thank you, :+1:

marcjansen:
I like it as well. Good to go IMO.

PR #75, opened 2016-04-14
This adds events before and after adding an external wms.

KaiVolland:


marcjansen:
:+1:, I'll merge this. A follow-up PR may provide API-docs.

PR #80, opened 2016-04-15
Fixed bindings for properties that have no setter method

weskamm:
Removed invalid bindings for `buttonText` and `emptyText` as these have no setter methods


marcjansen:
Looks good to me. I am going to merge this. Thanks @weskamm for noticing and fixing.

PR #81, opened 2016-04-19
Use additional configs to determine map component dynamically

annarieger:
With this PR some additional config properties  will be introduced to determine map component dynamically and get rid of hardcoded queries like `Ext.ComponentQuery.query('basigx-component-map')[0]`

This change is neccessary to be able to use `mapComponentConfig` object, if we like to use our own `xtype` of map component in the application.

S. here for example:
https://github.com/terrestris/BasiGX/blob/master/src/view/panel/MapContainer.js#L74.

These additional configs can be provided by instantiation of the corresponding `BasiGX` class in the application. Here example for `BasiGX.view.button.ZoomIn` class:

```js
{
      xtype: 'basigx-button-zoomin', 
      scale: 'small',
      olMap: olMap,
      ui: 'default'
}
```

### DIFF
--- a/src/view/button/Hsi.js
+++ b/src/view/button/Hsi.js
@@ -63,6 +63,12 @@ Ext.define("BasiGX.view.button.Hsi", {
     buttonPressed: true,
 
     /**
+     * Placeholder for the xtype of the map component (e.g. 'basigx-component-map').
+     * Will be used to be able to determine the map component dynamically
+     */
+    mapPanelXType: null,
+
+    /**
      *
      */
     config: {
@@ -97,7 +103,13 @@ Ext.define("BasiGX.view.button.Hsi", {
      *
      */
     setControlStatus: function(status){
-        var mapComponent = Ext.ComponentQuery.query('basigx-component-map')[0];
+        var me = this,
+            mapComponent = Ext.ComponentQuery.query(me.mapPanelXType)[0];
+
+        //fallback
+        if (Ext.isEmpty(mapComponent)) {
+            mapComponent = Ext.ComponentQuery.query('basigx-component-map')[0];
+        }
         mapComponent.setPointerRest(status);
     }
 });
--- a/src/view/button/Hsi.js
+++ b/src/view/button/Hsi.js
@@ -24,7 +24,8 @@ Ext.define("BasiGX.view.button.Hsi", {
     extend: "Ext.Button",
     xtype: 'basigx-button-hsi',
     requires: [
-        'Ext.app.ViewModel'
+        'Ext.app.ViewModel',
+        'BasiGX.util.Map'
     ],
 
     /**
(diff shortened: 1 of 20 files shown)


src/view/button/Hsi.js
```diff
@@ -97,7 +103,13 @@ Ext.define("BasiGX.view.button.Hsi", {
      *
      */
     setControlStatus: function(status){
-        var mapComponent = Ext.ComponentQuery.query('basigx-component-map')[0];
+        var me = this,
+            mapComponent = Ext.ComponentQuery.query(me.mapPanelXType)[0];
```
weskamm: This needs an additional check, as mapPanelXType may be null or undefined, which will cause an error here.
We could also use the config for that property and call `me.getMapPanelXType()`, check if it is defined and not empty and the proceed with ComponentQuery

src/view/button/ToggleLegend.js
```diff
@@ -52,17 +52,30 @@ Ext.define("BasiGX.view.button.ToggleLegend", {
...
         handler: function(button){
-            // TODO refactor so this works even outside of the mapcontainer
-            var legendPanel = button.up("basigx-panel-mapcontainer")
-                .down('basigx-panel-legendtree');
-            if(legendPanel.getCollapsed()) {
-                legendPanel.expand();
-            } else {
-                legendPanel.collapse();
+            var me = this,
+                legendPanel = Ext.ComponentQuery.query(me.legendPanelXType)[0];
```
weskamm: same as above

weskamm:
Is there a reason why you dont use the config object for the new property? And also a utility method would make sense to try to get the map, so we can extend / fix that part at one point in the future if needed.

Please adress the mentioned issues above with the null check, then this is ok for me :+1:

src/view/button/ToggleLegend.js
marcjansen: also the `up().down()`-logic, while looking strange, is probably less error-prone than a general component query (which I hat, you all know that). It can be made more flexible by using the provided `legendPanelXType`, of course.

annarieger:
Thanks for your hints, @weskamm! I've refactored some classes now to get more dynamic by determining of `mapComponent` object.

Also utility methods `getMapComponent()` and `getLegendTreePanel()` were introduced in the class `BasiGX.util.Map`.

It would be great, if you could review the changes once again, please.

src/view/container/Redlining.js
```diff
@@ -209,19 +214,20 @@ Ext.define("BasiGX.view.container.Redlining", {
...
-       var mapComponent = Ext.ComponentQuery.query('gx_component_map')[0];
-       var map = mapComponent.getMap();
        var displayInLayerSwitcherKey = BasiGX.util.Layer.
            KEY_DISPLAY_IN_LAYERSWITCHER;
 
-       if (!me.redliningVectorLayer) {
+       //set map
+       me.map = BasiGX.util.Map.getMapComponent().getMap();
+
+       if (!me.redlmapiningVectorLayer) {
```
weskamm: i think there is a map where it doesn't belong :-)

weskamm:
Just a minor issue left, this is looking really good! Nice to have a util for this now.
Feel free to merge after adressing the last issue :+1:

src/view/container/Redlining.js
annarieger: Oooh, what a nice one :innocent: 

Fixed it in 05d9344.

annarieger:
Thanks a lot, @weskamm! Merging now..

PR #84, opened 2016-04-20
Hiding the coordinate transform vector layer as default

weskamm:
title says it all


dnlkoch:
We should think about having the layer as config option in the future, but for the moment I'm more than +1 !

weskamm:
Thanks, merging...

PR #85, opened 2016-04-21
BugFix for LegendTree-RowExpander

KaiVolland:
This fixes a bug that removes items mistakenly.

This part of code searches for an item that is added twice. And removes it before. The selection of the item wasn't specfic enough so to many items where removed. Now the items is selected more specifcly.


marcjansen:
Looking great for now, @KaiVolland. We both know this needs a refactoring.

Thanks for this!

Please merge.

PR #86, opened 2016-04-22
Add methods to collapse/expand/toggle legends

marcjansen:
We previously had a method `expandAllBodies`, that would throw an exception
when being called (it called an not-existing method). It also wouldn't expand
all bodies, but merely toggle all of them.

This PR adds the following methods:

* `collapseAllBodies()`,
* `expandAllBodies()`,
* `toggleAllBodies()`.

Please review.

/cc @KaiVolland

### DIFF
--- a/src/view/panel/LegendTree.js
+++ b/src/view/panel/LegendTree.js
@@ -136,15 +136,76 @@ Ext.define("BasiGX.view.panel.LegendTree", {
     },
 
     /**
-     * Expands All RowBodies
+     * Expands, collapses or toggles all row bodies with the components,
+     * depending on the passed mode.
+     *
+     * @param {String} mode The mode to set; either `toggle`, `expand` or
+     *     `collapse`.
+     * @protected
+     */
+    setModeAllBodies: function(mode) {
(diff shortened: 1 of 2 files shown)


src/view/panel/LegendTree.js
```diff
@@ -136,15 +136,77 @@ Ext.define("BasiGX.view.panel.LegendTree", {
...
+                ' to be bei either "toggle", "expand" or "collapse".');
+            return;
+        }
         var me = this;
-        if((!me.getBodyInitiallyCollapsed()) && me.plugins.length > 0){
-            me.getStore().each(function(record, index) {
-                me.plugins[0].toggleRow(index, record);
-            });
+        // TODO below detection could be enhanced
+        var plugin = me.plugins.length > 0 ? me.plugins[0] : null;
```
KaiVolland: Maybe you can use getPlugin()

KaiVolland:
Just one minor.

marcjansen:
Cheers for the review and hint, @KaiVolland. I added and used the `pluginId` in 	d5f1c51.

I'll merge once checks are green.

PR #88, opened 2016-05-11
Show correct initial scale value on first application load

annarieger:
This PR fixes the shown scale value in `ScaleCombo`, that will initially be shown on the first application load.

Previously ths `resolution` value was shown instead of `scale`.

Please review.

/cc @weskamm

### DIFF
--- a/src/view/combo/ScaleCombo.js
+++ b/src/view/combo/ScaleCombo.js
@@ -126,7 +126,9 @@ Ext.define("BasiGX.view.combo.ScaleCombo", {
         me.callParent([arguments]);
 
         // set the correct default value
-        me.setValue(me.map.getView().getResolution());
+        var defaultScale = Math.round(me.getCurrentScale(
+                me.map.getView().getResolution())).toLocaleString();
+        me.setValue('1:' + defaultScale);
 
         // register listeners to update combo and map
         me.on('select', function(combo, rec) {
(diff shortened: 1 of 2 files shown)


weskamm:
I don't think that it is a bug. You need to make sure, that your application uses the same resolutions as you have defined in the creation of the scale combo.
Example:



``map.getView().setResolution(560);``

``Ext.create('Ext.window.Window', {``
``    width: 200,``
``    height: 200,``
``    items: [{``
``        xtype: 'basigx-combo-scale',``
``        cls: 'basigx-combo-scale',``
``        scales: [``
``                         {"scale": "1:2.000.000", "resolution": 560},``
``                         {"scale": "1:1.000.000", "resolution": 280},``
``                         {"scale": "1:500.000", "resolution": 140},``
``                         {"scale": "1:250.000", "resolution": 70},``
``                         {"scale": "1:100.000", "resolution": 28},``
``                         {"scale": "1:50.000", "resolution": 14},``
``                         {"scale": "1:25.000", "resolution": 7},``
``                         {"scale": "1:10.000", "resolution": 2.8},``
``                         {"scale": "1:5.000", "resolution": 1.4},``
``                         {"scale": "1:2.500", "resolution": 0.7},``
``                         {"scale": "1:1.000", "resolution": 0.28}``
``           ]``
``    }]``
``}).show();``

This should show the scale of 1:2.000.000, as the resolution of the map now fits this scale.
Setting the value with resolution is fine, as we use scale as `displayfield`

annarieger:
Ok, I faced out the problem with my application. The set of map resolutions doesn't contain any of predefined values of `scales`-array of the `basigx-combo-scale` component. That's why `me.setValue(me.map.getView().getResolution());` call can't set the corresponding mapped scale value to the current resolution got from the map.

To avoid this problem, we must be sure here, that the set of predefined map resolutions and scale-config of `basigx-combo-scale` contain the same values.

This PR won't be merged, I'll close it here and provide the adjusted config object in my project instead.

Thx for review and support, @weskamm! :crown:

PR #94, opened 2016-05-24
Fixed buttons to be mobile compatible again

weskamm:
This PR contains a change that has been removed by accident in a previous commit.
We still need to check if the `setTooltip` function is available before we bind the value, as in fact a tooltip for a button does not exist in the modern toolkit.


marcjansen:
Plus 1. Please merge.

weskamm:
thanks @marcjansen

PR #96, opened 2016-06-01
Added WMS-time slider

mmacata:
This slider takes a WMS-Time Layer and uses its timestamps to create itself. This makes it possible to slide through the different times from the layer.

The tests depend on an ajax request to a public WMS-time. If this foreign depency is not good practice it can be removed.

### DIFF
--- a/src/view/container/WMSTimeSlider.js
+++ b/src/view/container/WMSTimeSlider.js
@@ -1,0 +1,196 @@
+/* Copyright (c) 2016 mundialis GmbH & Co. KG
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+/**
+ * WMS-Time Slider
+ *
+ * Used to slide through all timestamps from a WMS-Time Layer. Timestamps are
+ * taken from the layer's getCapabilitites document when displayed as a list.
+ *
+ * Example usage:
+ *
+ * {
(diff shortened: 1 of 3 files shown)


src/view/container/WMSTimeSlider.js
```diff
@@ -0,0 +1,196 @@
...
+        var params = Ext.apply({
+            service: 'WMS',
+            version: '1.3.0',
+            request: 'GetCapabilities'
+        });
+        Ext.Ajax.request({
+            url: slidercontainer.url,
+            method: 'GET',
+            params: params,
+            async: false,
```
weskamm: we should try to avoid this and use events to handle the asynchronous answer

src/view/container/WMSTimeSlider.js
```diff
@@ -0,0 +1,196 @@
...
+                    }
+                }
+            },
+            {
+                xtype: 'slider',
+                name: 'timeslider',
+                tipText: function(thumb){
+                    var slider = Ext.ComponentQuery.query('basigx-container-wmstimeslider')[0];
+                    var time = slider.getTimes()[thumb.value];
+                    return Ext.Date.format(new Date(time), 'Y-m-d');
```
weskamm: i know its already a todo, but having the timeformat configurable would be nice!

weskamm:
added minor comments, nice component!

src/view/container/WMSTimeSlider.js
```diff
@@ -0,0 +1,196 @@
+/* Copyright (c) 2016 mundialis GmbH & Co. KG
```
marcjansen: I think this needs to be changed to `terrestris`

src/view/container/WMSTimeSlider.js
```diff
@@ -0,0 +1,196 @@
...
+        times: []
+    },
+
+    url: null,
+
+    layerNameInWMS: null,
+
+    layerNameInClient: null,
+
+    cls: 'timeSlider',
```
marcjansen: We usually have a prefix for our classes, at least I think that we normally use `basigx-`

src/view/container/WMSTimeSlider.js
marcjansen: Yes, I agree with @weskamm

KaiVolland:
@mmacata Are you going to tackle this one or can we close it?

KaiVolland:
I'll close this now. Feel free to reopen if this still an actual PR.
@mmacata

PR #97, opened 2016-07-06
replaces MapQuest layer source

mmacata:
In ol 3.17.0 the source MapQuest was removed. See https://github.com/openlayers/ol3/releases/tag/v3.17.0 . So it would be better to use a different source in the overviewmap.

### DIFF
--- a/src/view/panel/MapContainer.js
+++ b/src/view/panel/MapContainer.js
@@ -102,7 +102,7 @@ Ext.define("BasiGX.view.panel.MapContainer", {
             layers: [
                 new ol.layer.Tile({
                       source: new ol.source.TileWMS({
-                          url:"htto://ows.terrestris.de/ows?",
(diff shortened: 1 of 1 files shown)


src/view/panel/MapContainer.js
```diff
@@ -101,7 +101,10 @@ Ext.define("BasiGX.view.panel.MapContainer", {
             hidden: true,
             layers: [
                 new ol.layer.Tile({
-                    source: new ol.source.MapQuest({layer: 'sat'})
+                      source: new ol.source.TileWMS({
+                          url:"htto://ows.terrestris.de/ows?",
```
annarieger: typo `htto` instead of `http`

weskamm:
besides, the url should point to http://ows.terrestris.de/osm/service? instead

mmacata:
you are right, url is fixed now.

weskamm:
thanks, merging this now...

PR #99, opened 2016-07-21
Several redlining enhancements

weskamm:
This PR fixes several minor redlining issues.
The usability has improved, as the interactions now make use of a hovereffect, indicating which feature is currently selected to modify / delete etc.

The text of postits is now editable with the modify tool.

The util.Animate has been made compatible for changes in ol3 later than v3.15


dnlkoch:
Toptop!

weskamm:
Thanks for review, merging now

PR #100, opened 2016-08-09
Be more cautious when addding a filter

marcjansen:
In an application I observed that the `layer` from `getOlLayer` might be `undefined` (I issued an unlikely but allowed `removeAll` on the store of the layer tree). 

We should guard against this when adding a filter to our legend panel.

Please review.


dnlkoch:
Great! Thanks for this, I'll merge.

marcjansen:
Cheers for the review, @dnlkoch.

PR #101, opened 2016-08-09
Give the rowexpander plugin a pluginId

marcjansen:
This way the plugin instance can be retrieved with code like

```js
    legendTree.getPlugin('rowexpanderwithcomponents');
```

Previously there was no way except for iterating over the return value
of `legendTree.getPlugins()`.


dnlkoch:
Great. Thanks for this, I'll merge.

marcjansen:
Cheers for the review, @dnlkoch.

PR #102, opened 2016-08-29
Removed gx_symbolizercolumn due to removal in GeoExt3

weskamm:
These changes are needed due to the removal of the gx_symbolizercolumn in GeoExt3

### DIFF
--- a/src/view/container/NominatimSearch.js
+++ b/src/view/container/NominatimSearch.js
@@ -33,7 +33,7 @@ Ext.define("BasiGX.view.container.NominatimSearch", {
 
     requires: [
         'GeoExt.data.store.Features',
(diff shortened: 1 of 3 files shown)


marcjansen:
Have you had a chance to test these? There is an open bug somewhere...

If it works in one pf our apps I am confident this can be merged.

chrismayer:
If you discover problems due to the mentioned bug (https://github.com/geoext/geoext3/issues/171) you could remove the explicit ``fields`` declaration from your store as a temp. workaround (while the bug isn't fixed yet).

weskamm:
i have tested one of the three components and its working fine. As the changed code is the same in all three, i think this will be ok

weskamm:
Thanks for review, merging now

PR #105, opened 2016-09-29
Style enhancements

dnlkoch:
In this PR you'll find:
* A new method `rgbaToHex8()` to convert a RGBA color string into an hex8 color string.
* A default (and empty) `ol.Style.Text` object in any redlining style.


marcjansen:
Thanks, please merge when green.

PR #107, opened 2016-10-06
Print form fixes

annarieger:
This MR provides two fixes for `BasiGX.view.form.Print` class:

1. `northArrow` attribute definition was unusable due to wrong object assignaiton
2. Method `renderAllClientInfos` will be called much more rarely now since a `moveend` event of the map will be used instead of too global `changeproperty` event of the ol.View. Additionally this event will be unregistried again after print window was closed.

Please review @weskamm


annarieger:
Thank you, @weskamm! :pineapple:

PR #110, opened 2016-11-10
Update animate util

mmacata:
This PR updates the animate util as it uses the ol3 methods on the vector context which changed in v3.15.0 (https://github.com/openlayers/ol3/releases/tag/v3.15.0).

It can be still used with older versions.


weskamm:
nice one, i thought this had been fixed already but looks like im wrong :-)
merging now...

PR #112, opened 2016-11-22
Warning in case of text too long for postits

weskamm:
This change displays a configurable message when the maxLength of the text for postits has been exceeded. The user can then continue or edit the text


marcjansen:
Nice thanks, please merge.

PR #114, opened 2016-11-22
Making the postit work better with screen share

weskamm:
This enables clean usage of screen sharing when using postits, as the style of a postit has to be set before it gets pushed to the redlinefFeatures collection. Also enhanced the modifyPostit function to reflect latest changes on maxLength


dnlkoch:
Check!

weskamm:
Thanks, merging

PR #119, opened 2016-12-05
Allow dragging and resizing of print extent

marcjansen:
This PR suggests to add the possoibility to drag and resize the print extent dynamically.

We use the excellent `ol.interaction.Transform` from https://github.com/Viglino/ol3-ext for the interactivity of the extent feature.

This is what is in this PR:

* A utility class `BasiGX.ol3.extension.TransformInteraction` that, when required, will set up the interaction class `ol.interaction.Transform` in the global namespace. The code is taken directly from the linked project, and basically has been reformatted. See the file-header for details on changes.

* Two new utilities in `BasiGX.util.Map`: `getResolutionForScale` and `getScaleForResolution`

* Three new configs for `BasiGX.view.form.Print`: `printExtentAlwaysCentered`, `printExtentMovable`, `printExtentScalable`. The default have been chosen to be backwards compatible. If you want to enable the dynamic capabilities, you need to turn of `printExtentAlwaysCentered` and the enable any combination of `printExtentMovable` & `printExtentScalable`.

* The print form class also got more comments and was restructured a bit (functions that are called during initialisation e.g.)

Future PRs should handle the rotation in a sane manor, the needed functionality is already included in the new interaction.

Please review.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,9 +34,10 @@
     },
     "output": "${package.dir}/build",
     "scripts": {
-        "pretest": "npm run lint-js",
+        "pretest": "npm run lint",
         "test": "mocha-phantomjs --ignore-resource-errors --ssl-protocol=any --ignore-ssl-errors=true --local-to-remote-url-access=true test/index.html",
-        "lint-js": "eslint -c .eslintrc src/ && eslint -c test/.eslintrc test/",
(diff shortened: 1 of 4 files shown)


buehner:
+1 from me :wink:

marcjansen:
Cool, thanks for the reviews.

PR #121, opened 2016-12-07
Add missing msg field for angle measurement

buehner:
This PR adds a missing viewModel field declaration that is required (as a default value) to show the appropriate text when drawing an angle.


weskamm:
:+1: @buehner , :-1: @marcjansen

PR #122, opened 2016-12-07
Hover plugin minor bugfixes

dnlkoch:
Here we cover two minor issues regarding the `Hover` plugin:
* Introduce `enableHoverSelection` to control wether to allow click selection of a hovered feature.
* Only remove ol `Overlays` that are controlled by the plugin itself.


weskamm:
:+1: looking good to me!

buehner:
:+1:

PR #123, opened 2016-12-13
Check locale config before applying

marcjansen:
This PR suggest to guard against locale configs that look misconfigured, and effectively allows us to do things like the following:

```js
// in a button controller
Ext.create('Ext.window.Window', {
     viewModel: myBtnViewModel, // a real instance of the vm
     bind: {
         title: '{myWinTitle}' // defined key in btn viewmodel
     }
     // other win properties
});
```

That particluar window will benefit from changed locales (with resulting changed viewModels for instantiated things) as long as the original button is rendered / instanciated.

Please review.

/cc @annarieger & @dnlkoch


annarieger:
Thanks for this addition, @marcjansen :+1: 

Please merge.

marcjansen:
Thanks for the review, @annarieger.

PR #126, opened 2017-01-05
Make the add wms window configurable

buehner:
This makes the "Add WMS" window (optionally) configurable by passing values for the window in the `windowConfig` object that is introduced in this PR.

Existing applications do not need any adaptions (i.e. this code should work fine for them, too).

### DIFF
--- a/src/view/button/AddWms.js
+++ b/src/view/button/AddWms.js
@@ -74,7 +74,7 @@ Ext.define("BasiGX.view.button.AddWms", {
                     }]
                 };
 
-                var windowConfigToApply = this.getConfig('windowConfig');
+                var windowConfigToApply = this.getWindowConfig();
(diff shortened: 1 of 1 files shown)


src/view/button/AddWms.js
```diff
@@ -67,7 +72,12 @@ Ext.define("BasiGX.view.button.AddWms", {
                     items: [{
                         xtype: 'basigx-form-addwms'
                     }]
-                }).show();
+                };
+
+                var windowConfigToApply = this.getConfig('windowConfig');
```
marcjansen: Out of curiosity, why don't you use `this.getWindowConfig()`?

buehner: Because i didn't know it better :wink: 
I changed this call as it seems to be smarter! Thx for the hint...

PR #127, opened 2017-01-13
Improves opacity support for redline styler.

KaiVolland:


dnlkoch:
Thanks, great addition!

PR #130, opened 2017-01-17
Make custom handling for initially collapsed panel version dependent

dnlkoch:
See in title (and comment).


marcjansen:
:+1: cool we have had the right gut feeling (see logging notice). Nice fix, thanks @dnlkoch and @KaiVolland

PR #132, opened 2017-01-20
Introduces copy to clipboard button

KaiVolland:
All credits go to @dnlkoch. <3

:hankey: with :heart: by terrestris


src/util/CopyClipboard.js
```diff
@@ -0,0 +1,56 @@
+Ext.define('BasiGX.util.CopyClipboard', {
```
marcjansen: No API docs :disappointed:

src/util/CopyClipboard.js
```diff
@@ -0,0 +1,56 @@
...
+
+    statics: {
+
+        /**
+         *
+         */
+        copyToClipboardSupported: document.queryCommandSupported('copy'),
```
marcjansen: Does IE or Edge have `queryCommandSupported`?

dnlkoch: Yes, see [here](https://developer.mozilla.org/en-US/docs/Web/API/Document/queryCommandSupported#Browser_compatibility).

src/util/CopyClipboard.js
dnlkoch: You're so right. Added in 7ac3c0f4.

PR #133, opened 2017-01-25
Make LayerSetChooser configurable with local config JSON

chrismayer:
This introduces the possibilty to configure a ``BasiGX.view.panel.LayerSetChooser`` by a local config JSON-object, which is passed and interpreted to the underlying ``BasiGX.view.view.LayerSet``.

### DIFF
--- a/src/view/panel/LayerSetChooser.js
+++ b/src/view/panel/LayerSetChooser.js
@@ -84,6 +84,13 @@ Ext.define("BasiGX.view.panel.LayerSetChooser", {
     layerSetUrl: null,
 
     /**
+     * JSON-object to be passed to the automatically created
+     * 'basigx-view-layerset'.
+     * Only has an effect if this.items is empty.
+     */
+    layerSetData: null,
+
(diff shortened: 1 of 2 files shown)


src/view/panel/LayerSetChooser.js
```diff
@@ -84,6 +84,13 @@ Ext.define("BasiGX.view.panel.LayerSetChooser", {
     layerSetUrl: null,
 
     /**
+     * JSON-object to be passed to the automatically created
+     * 'basigx-view-layerset'.
+     * Only has an effect if this.items is empty.
+     */
+    layerSetdata: null,
```
marcjansen: Can we name it `layerSetData` (capital *`D`*)? That would match `layerSetUrl` …

chrismayer: Actually it was my intention to write it with ``D``, missed it somehow... i'll fix it.

marcjansen:
Thanks for the change!

PR #135, opened 2017-02-01
Build using latest sencha / ExtJS during CI

marcjansen:
This PR suggests to upgrade to using ext 6.2.0 and the newest sencha cmd (v6.2.1.29).

Please review.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ build/
 sencha-error-*.log
 bootstrap.json
 sass/example/bootstrap.json
+sass/example/bootstrap.jsonp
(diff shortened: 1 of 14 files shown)


marcjansen:
I have no idea why the build fails here… investigating now

marcjansen:
Since the build is green now, I think this is ready to be reviewed. This is basically the same we had for GeoExt: https://github.com/geoext/geoext3/pull/193

annarieger:
Looks very good to me, @marcjansen! Thanks for upgrading :+1

marcjansen:
Thanks for the review!

PR #136, opened 2017-02-06
Add the ability to add markup to enhance accessibility

marcjansen:
This adds three components that will enhance the accessibility of apps created using them:

* A plugin to add the title of a component as hidden header (e.g. in a `<h2>`-tag): `BasiGX.plugin.AccessibleTitle`
  * That plugin will either render to the `header` or the `body` element of its target
  * The plugin also overides the `setTitle` method to update the hidden HTML accordingly when the title changes.
* Two base classes that use this plugin:
  * `BasiGX.view.panel.Accessible`: will render their `title` into a `<h3>`
  * `BasiGX.view.window.Accessible`: will render their `title` into a `<h2>`
  * Both the panel and the window class do this even when configured with `header: false`
  * A follow up PR might make the panel- class the base class for all our BasiGX panels

Please review.

### DIFF
--- a/src/plugin/AccessibleTitle.js
+++ b/src/plugin/AccessibleTitle.js
@@ -111,7 +111,7 @@ Ext.define('BasiGX.plugin.AccessibleTitle', {
         if (!headerDom) {
             return;
         }
-        me.addedHtmlHeader = Ext.DomHelper.append(headerDom, spec);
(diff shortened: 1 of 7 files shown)


dnlkoch:
Great one!

marcjansen:
Thanks for the review, @dnlkoch

PR #137, opened 2017-02-06
Add and use a button base class

marcjansen:
This PR suggests to add a common base class for the buttons which allows us to stop copy and paste the constructor part where we have special logic wrt to `setTooltip`.

@annarieger likes it already :smile: but I'd like to as @dnlkoch to look into this as well (in the project he currently working on…)

### DIFF
--- a/src/view/button/Base.js
+++ b/src/view/button/Base.js
@@ -1,0 +1,40 @@
+/* Copyright (c) 2017-present terrestris GmbH & Co. KG
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+/**
+ * A base class for buttons.
+ *
+ * @class BasiGX.view.button.Base
+ */
+Ext.define('BasiGX.view.button.Base', {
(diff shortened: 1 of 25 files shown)


src/view/button/Base.js
```diff
@@ -0,0 +1,38 @@
+/* Copyright (c) 2015 terrestris GmbH & Co. KG
```
annarieger: Wrong year

annarieger:
Thanks for this simplification, @marcjansen! :pineapple: 

Please adjust a year in license agreement and make travis happy again.

src/view/button/Base.js
```diff
@@ -0,0 +1,38 @@
...
+     *
+     */
+    constructor: function() {
+        var me = this;
+        me.callParent(arguments);
+
+        if (me.setTooltip) {
+            var bind = me.config.bind;
+            bind.tooltip = me.getViewModel().get('tooltip');
+            me.setBind(bind);
```
chrismayer: Is it necessary to explicitly do the binding of the tooltip? Isn't this done by the ExtJS MVVM-mechanism?

marcjansen: It usually is, Previously this was done by the subclasses. We need this special handling so we can use the button (view) classes for both the `modern` and `classic` toolkit. The tollkit `modern` misses `setTooltip`, which is why we have handled it like this in all the button classes.

src/view/button/Base.js
marcjansen: Already adjusted, But I want to test this first in real applications, before I will merge this.

src/view/button/Base.js
chrismayer: Oh, really good point! Thanks for the detailed explanation.

marcjansen: Actually I was thinking about this a bit more. It looks as if we could have this done differently, at lesst with 6.2.0. I might come up with a different approach for the modern toolkit... But for now this is only a refactoring of the existing approach to handle both modern and classic.

Thanks for the comment!

marcjansen:
Thanks @annarieger,  @dnlkoch and @chrismayer for the valuable feedback and review.

PR #138, opened 2017-02-07
Print form layout fixes/enhancements

annarieger:
Minor layout fixes for print form:
* use `buttons` property instead of `bbar` to harmonize CSS outline with other BasiGX components
* use `anchor` layout for fieldset to fix missed/non uniform padding
* added some labels to enhance usability

Please review @terrestris/owners

### DIFF
--- a/examples/i18n/resources/locale/app-locale-de.json
+++ b/examples/i18n/resources/locale/app-locale-de.json
@@ -228,7 +228,12 @@
                 "printButtonSuffix": "anfordern",
                 "downloadButtonPrefix": "Download",
                 "printFormat": "pdf",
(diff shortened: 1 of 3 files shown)


marcjansen:
We have to check this with existing applications that use the PrintForm. It'd be extra cool if @KaiVolland or / @dnlkoch could test whether this breaks their project.

I am very cautious here, because a simple change in this class broke it some time ago.

Can we put this on hold until they give their :+1: ? What do you think @annarieger?

dnlkoch:
Verified. Please merge at will.

annarieger:
Cheers for review, guys! :pineapple:²

Merging now..

PR #144, opened 2017-02-10
Add a focusable tree item class

marcjansen:
This PR suggests to add a class that can be used inside of `Ext.list.Tree` instances. This tree item class differs to its parent in two points:

* `BasiGX.view.list.FocusableTreeItem` instances can be focused; by clicking or by hitting the tab-key.
* If a focused treeitem receives a `keypress`-event that originates from either the space-bar or the the enter-key, the `selectionchange` event of the parent `Ext.list.Tree` is fired with the focused item passed as new selection.

Please review.


annarieger:
And once again - great one! Thanks @marcjansen :pineapple:

marcjansen:
Thanks for the review, Anna.

PR #156, opened 2017-03-08
Add overviewmapToggle event

KaiVolland:
This adds an event when the overviewmap is toggled. This helps to react to changes it the overviewmaps visibility.


marcjansen:
Thanks for the doc change

PR #161, opened 2017-03-17
added hover revert feature order

mpochert:
Hallo terrestris,
da die Sortierreihenfolge bei getFeatureInfo() zwischen Geometriedarstellung und HoverInfo differiert, soll ein Layer die Möglichkeit besitzen die Sortierreihenfolge im HoverInfo-Fenster umkehren zu können. Dann stimmt die Reihenfolge im HoverInfo-Fenster mit der "Render-Reihenfolge" der Features überein.
Wäre super wenn Ihr die Änderung in die BasiGX übernehmen könntet.

Viele Grüße
Marco Pochert

### DIFF
--- a/src/plugin/Hover.js
+++ b/src/plugin/Hover.js
@@ -35,6 +35,14 @@ Ext.define('BasiGX.plugin.Hover', {
         LAYER_HOVERABLE_PROPERTY_NAME: 'hoverable',
 
         /**
+         * The property of a layer that holds a boolean value which indicates
+         * whether this layer features shall be reverted on hover tooltip.
+         *
+         * @type {String}
+         */
+        LAYER_HOVER_FEATURES_REVERT_NAME: 'hoverFeaturesRevert',
+
+        /**
          * The property of a layer that holds a string value which indicates,
          * which field of the layer shall be shown when hovering.
          *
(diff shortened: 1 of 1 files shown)


marcjansen:
@mpochert Thanks for this contribution, looks good at first sight. I am not so sure if I like the name of the property, but cannot come up with sth. better right now.

/cc @KaiVolland @dnlkoch @akohlrusch: Do you think you can test this in an application? TIA

marcjansen:
Ping @KaiVolland @dnlkoch @akohlrusch Any chance someone can look into this?

KaiVolland:
I definetly see the problem detected by @mpochert.

Thanks for the PR but I think it's a bug and should not be configurable.

I tested it a few times and the order (z-index) of the hoverfeatures seems to be somewhat random or at least not always the same for the same to features. So reverting the features would not solve the problem.

I think the hover-features should always represent the current order of the layers.

mlechner:
According to JSON specs "An array is an ordered collection of values". Therefore the order of the features in a getfeatureinfo request has to be reproducible (and in our testcases is as a result of the order by statement in the sql view definition)
The real problem here is that the order of a requested attribute table usually is reverse the order needed for a hover popup. In an attribute table the most recent records are at the bottom while in a popup it might be more appropriate to revert the order to get the most recent records to top.
Making it configurable seems to be the appropriate way to do it like this (whithout the necessity to implement client side order-functions)
@KaiVolland please call back to discuss this if still unclear!

marcjansen:
Ping everyone, what needs to be done to get this resolved?

weskamm:
So the conclusion is, that this new feature makes sense. I think the naming is ok and since this issue is open for so long, i will merge it now as is.
Thanks everybody!

PR #163, opened 2017-03-30
Multiple enhancements of the AddWms form panel

marcjansen:
This PR suggests some changes to the `AddWms` form:

* Rename the configuration `versionsWmsAutomatically` to `autoDetectVersion`
* Remove configuration `triedVersions`, and instead use an internal property as this should never be set from the user
* Change the `defaultUrl`s default value to be the https-variant
* Add configuration `useDefaultXhrHeader` (to be able to communicate with more actual WMS servers out there, see doc comments for details)
* Add configuration option `disableCaching` (Named as the `Ext.Ajax.request` configuration) to control whether the `_dc`-parameter shall be sent along when fetching the capabilities
* Better error handling:
  * OGC `ServiceExceptionReports` are reported with details
  * Timeouts  are detected and reported
  * Some very common http status codes get special error messages
  * CORS errors are also detected as good as possible
  * All error messages live in the `viewModel`

The first two bullet points make this PR backwards incompatible, so think twice before merging, client projects may need to be updated as follows:

  * When using the former configuration option `versionsWmsAutomatically`: Rename this option to `autoDetectVersion`
  * (Very unlikely) When using the former configuration option `triedVersions`: Remove this option.

### DIFF
--- a/src/view/form/AddWms.js
+++ b/src/view/form/AddWms.js
@@ -56,10 +56,27 @@ Ext.define('BasiGX.view.form.AddWms', {
             addCheckedLayersBtnText: 'Ausgewählte Layer hinzufügen',
             errorIncompatibleWMS: 'Der angefragte WMS ist nicht kompatibel ' +
                     'zur Anwendung',
-            errorRequestFailed: 'Die angegebene URL konte nicht abgefragt ' +
+            errorRequestFailed: 'Die angegebene URL konnte nicht abgefragt ' +
                     'werden',
             errorCouldntParseResponse: 'Die erhaltene Antwort konnte nicht ' +
-                    'erfolgreich geparst werden'
+                    'erfolgreich geparst werden',
+            msgRequestTimedOut: 'Die Anfrage wurde nicht schnell genug ' +
+                    'beantwortet und abgebrochen',
+            msgServiceException: 'Eine OGC ServiceException ist aufgetreten',
+            msgCorsMisconfiguration: 'HTTP access control (CORS) auf dem ' +
+                    'Zielserver vermutlich nicht korrekt konfiguriert',
+            msgUnauthorized: 'Der Client ist nicht gegenüber dem Zielserver ' +
+                    'authentifiziert',
+            msgForbidden: 'Der Client hat keinen Zugriff auf die angefragte ' +
+                    'Ressource',
+            msgFileNotFound: 'Die angefragte Ressource existiert nicht',
+            msgTooManyRequests: 'Der Client hat zu viele Anfragen gestellt',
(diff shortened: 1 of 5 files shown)


marcjansen:
Thois is ready for review. I hop to add test oin the upcoming week

/cc @KaiVolland & @dnlkoch

src/view/form/AddWms.js
```diff
@@ -56,10 +56,27 @@ Ext.define('BasiGX.view.form.AddWms', {
...
+            msgForbidden: 'Der Client hat keinen Zugriff auf die angefragte ' +
+                    'Ressource',
+            msgFileNotFound: 'Die angefragte Ressource existiert nicht',
+            msgTooManyRequests: 'Der Client hat zu viele Anfragen gestellt',
+            msgServiceUnavailable: 'Der Server ist derzeit nicht verfügbar ' +
+                    '(zu viele Anfragen bzw. Wartungsmodus)',
+            msgGatewayTimeOut: 'Der Server fungierte als Gateway und das ' +
+                    'Originalziel hat zu langsam geantwortet',
+            msgClientError: 'Ein unspezifizierter Clientfehler ist aufgetreten',
+            msgServerError: 'Ein unspezifizierter Serverfehler ist aufgetreten'
```
annarieger: IMHO, all these strings should be added to the locales as well:
https://github.com/terrestris/BasiGX/blob/01b0e45c4aa24d67da4f4dbeb176621a75522ec4/examples/i18n/resources/locale/app-locale-de.json#L189
https://github.com/terrestris/BasiGX/blob/01b0e45c4aa24d67da4f4dbeb176621a75522ec4/examples/i18n/resources/locale/app-locale-en.json#L189

src/view/form/AddWms.js
```diff
@@ -289,99 +345,174 @@ Ext.define('BasiGX.view.form.AddWms', {
...
+            return 'msgServiceUnavailable';
+        } else if (status === 504) {
+            return 'msgGatewayTimeOut';
+        } else if (status >= 400 && status < 500) {
+            return 'msgClientError';
+        } else if (status >= 500) {
+            return 'msgServerError';
+        }
+        return null;
+    },
```
annarieger: Wow :+1:

annarieger:
Looks very good to me, especially the error handling is impressive :+1: 

Please check the backward compatibility against some other projects using this class and merge at will :pineapple:

marcjansen:
Thanks for the review, @annarieger:

* In order to add the i18n string to the example, I first had to fix it (c4ed8ee).
* In 59bef9b I then added the entries itself.

I will merge this once green.

PR #166, opened 2017-04-10
Fixes on graphicpool

weskamm:
This PR changes 2 things:
1.) Graphic Pool does not get closed anymore on succcesful image deletion
2.) Parsing of response on image deletion when using REST interface wwill not fail anymore as we check for empty responses and statuscode


weskamm:
thx!

PR #170, opened 2017-05-08
WIP:Always try to reload the graphic pool

buehner:
I tried using the `basigx-panel-graphicpool` component in a SHOGun2 based project (which is Multipart-Upload compatible, i.e. has an according `MultipartFilter ` (to handle CSRF) and a `CommonsMultipartResolver`)

At some point I recognized that the upload (using the graphic pool component) is working fine (HTTP status 200 with "JSON as string" success-response and text/html content type). Even though everything seems fine, i always end up in the `failure` function (changed within this PR).

The changes within this PR parse the response and try to handle it as a success (if that makes sense, i.e. HTTP Status 200 and a "valid" JSON).

I could not find a smarter solution, so this is marked as WIP. I think we should discuss this.

### DIFF
--- a/src/view/panel/GraphicPool.js
+++ b/src/view/panel/GraphicPool.js
@@ -226,8 +226,21 @@ Ext.define('BasiGX.view.panel.GraphicPool', {
                         BasiGX.error(me.getViewModel().get('uploadErrorText'));
                     }
                 },
-                failure: function() {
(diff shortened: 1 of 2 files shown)


weskamm:
is this still an issue or is it obsolete?

weskamm:
Closing due to inactivity

PR #173, opened 2017-05-15
Adds Geometry Util.

KaiVolland:
This adds a small geometry util. Maybe if you don't want to include jsts for just one operation.

For know it only contains a static equals method to compare to ol.geometries.

It could be extended with further functions.

### DIFF
--- a/src/util/Geometry.js
+++ b/src/util/Geometry.js
@@ -1,0 +1,44 @@
+/* Copyright (c) 2016-present terrestris GmbH & Co. KG
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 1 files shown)


src/util/Geometry.js
```diff
@@ -0,0 +1,44 @@
+/* Copyright (c) 2016-present terrestris GmbH & Co. KG
```
marcjansen: 2017?

src/util/Geometry.js
```diff
@@ -0,0 +1,44 @@
...
+ * @class BasiGX.util.Geometry
+ */
+Ext.define('BasiGX.util.Geometry', {
+
+    statics: {
+
+        /**
+         * Compares two ol.geom.Geometries.
+         *
+         * @param {ol.geom.Geometry} geometry1
```
marcjansen: Please change as follows

`@param {ol.geom.Geometry} geometry1 The first geometry for the comparison.`

src/util/Geometry.js
```diff
@@ -0,0 +1,44 @@
...
+         * @param {ol.geom.Geometry} geometry2
```
marcjansen: Please change as follows

`@param {ol.geom.Geometry} geometry2 The second geometry for the comparison.`

marcjansen:
Still looking good to me. Please merge once the lines are shorter and the lights are green.

PR #183, opened 2017-06-26
Added postprocess step for config parser.

hwbllmnn:
Added a postprocess step in the map component. If the function
postprocessMap is defined in your overridden/custom ConfigParser, it
will be called once the component is created (via the call to
callParent).

Please review.


KaiVolland:
Seems to be not critical. I'll merge it.

PR #185, opened 2017-06-28
Supporting LineStrings and Polygons in flashFeature method

weskamm:
This change adds support for animated LineStrings and Polygons.
Support for older version of OpenLayers (below 3.15) may be added in upcoming PR.

Please review


marcjansen:
Nice, looks good to me. IMO we don't need support for ol < 3.15.

Please merge at will

PR #190, opened 2017-07-06
Fix top-level API docs of language combo

marcjansen:

Before:
before

After:
after


marcjansen:
I'll merge this without proper review now as a doc only change.

PR #192, opened 2017-07-08
Add package-lock.json

marcjansen:
`npm` (since v5.x.x) generates this file, and it should be committed in repositories. See e.g. here for an explanation: https://stackoverflow.com/a/44210813

Even `npm` itself advises us to commit it:

```
npm notice created a lockfile as package-lock.json. You should commit this file.
```

Please review.


annarieger:
LGTM, please merge :pineapple:

PR #193, opened 2017-07-08
Correct the docs of TtfGlyphInspector

marcjansen:
Mainly the `@class` had a wrong value.

Please review


marcjansen:
Since this is a doc only change, I'll merge without proper review.

PR #198, opened 2017-07-21
Fix uncaught TypeError when adding external WMS

buehner:
When using a GetCapabilities document with a style element that does not provide a LegendURL element, an TypeError was caught.

This PR fixes this.


annarieger:
Works for me, please merge :pineapple:

PR #199, opened 2017-07-21
Do not check all external WMS by default

buehner:
When using the AddWms functionality with a service that provides dozens of layers, they are all checked by default and it's very annoying to deselect all layers you dont want to add.

In my eyes, you most likely want to add a small number of layers, so the default should be false, i.e. you will first have to check the layers you want to add to your map.


annarieger:
+1 from me

weskamm:
i like, will merge now!

PR #204, opened 2017-07-26
Detect all word characters in the web project base name

dnlkoch:
This allows the detection of word characters (a-z, A-Z, 0-9, underscore) in the extraction of the web project base name out of `window.location.pathname`.


dnlkoch:
Thanks for the review, will merge.

PR #208, opened 2017-08-08
Introduces configurable tooltip for measurebutton.

KaiVolland:
This introduces tooltip texts to the inline viewmodel.
It also changes the keys for the buttontext to camelcase.

All in all i don't think its a good idea to set the button text in the constructor. But as i think that we rely on this in other projects i'm not going to change this behaviour.

@terrestris/devs Plz review.


dnlkoch:
Top!

PR #209, opened 2017-08-09
Keep transformed feature when moving the map

hwbllmnn:

### DIFF
--- a/src/view/form/Print.js
+++ b/src/view/form/Print.js
@@ -449,20 +449,15 @@ Ext.define('BasiGX.view.form.Print', {
             rotate: false
         });
 
-        me.transformInteraction.on('translateend',
-            function(event) {
-                me.currentExtent = event.feature;
-                me.renderAllClientInfos();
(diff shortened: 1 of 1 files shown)


src/view/form/Print.js
```diff
@@ -449,6 +449,21 @@ Ext.define('BasiGX.view.form.Print', {
...
+                var newFeat = me.extentLayer.getSource().getFeatures()[0];
+                me.transformInteraction.select(newFeat);
+            });
+        me.transformInteraction.on('scaleend',
+            function(event) {
+                me.currentExtent = event.feature;
+                me.renderAllClientInfos();
+                var newFeat = me.extentLayer.getSource().getFeatures()[0];
+                me.transformInteraction.select(newFeat);
+            });
```
dnlkoch: May be moved to (shareable) named function.

dnlkoch:
Top! Just a minor comment.

src/view/form/Print.js
hwbllmnn: True, done.

PR #210, opened 2017-08-30
Caching util methods

buehner:
These methods may be helpful if you are using a GeoWebCache (integrated in a GeoServer).

I'll try to add some tests tomorrow.

### DIFF
--- a/src/util/Caching.js
+++ b/src/util/Caching.js
@@ -1,0 +1,129 @@
+/* Copyright (c) 2015-present terrestris GmbH & Co. KG
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+/**
+ * Caching Util
+ *
+ * Some methods that may be helpful in case of caching, e.g. with GeoWebCache
+ *
+ * @class BasiGX.util.Caching
+ */
+Ext.define('BasiGX.util.Caching', {
+
+    requires: [
+        'BasiGX.util.Map'
+    ],
+
+    statics: {
+        /**
+         * Call this function from the console (in one of your applications)
+         * and it returns config information that may be used to quickly setup a
+         * matching GWC gridset in the GeoServer.
+         *
+         * Note: If the extent is not passed to this function, the ol.view of
+         * the ol.map should have explicitly set an 'extent' property!
+         *
+         * @param {Number} tileSize The tile size in pixels
+         * @param {Array} extent Optional: The extent to use for the gridSet
+         * @return {Object} An object with information to quickly setup a GWC
+         *     gridset in GeoServer.
+         */
+        getGeoWebCacheConfig: function(tileSize, extent) {
+            var me = this;
+            var map = BasiGX.util.Map.getMapComponent().getMap();
+            var mapView = map.getView();
+
+            if (Ext.isEmpty(tileSize)) {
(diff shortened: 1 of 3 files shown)


src/util/Caching.js
```diff
@@ -0,0 +1,129 @@
...
+
+            return gwcConfig;
+        },
+
+        /**
+         * Computes the tile origin (top left) for a given extent
+         * (lower left/upper right bbox), maxResolution and tileSize that
+         * can be set as the origin on a ol.tilegrid.TileGrid instance to
+         * match a GWC gridset configured in a GeoServer. GetMap requests
+         * should use the parameter tiled=true to make use of the cache.
```
marcjansen: Can you add a not that this is especially important as OL want us to provide the tile origin on the upper left side? And also please hint that the resulting grid should be usable for WMTS and WMS sources (having a GWC-integration, e.g. with `TILED=true`)

src/util/Caching.js
```diff
@@ -0,0 +1,129 @@
...
+         *
+         * @param {Number} tileSize The tile size in pixels
+         * @param {Number} maxResolution The max resolution
+         * @param {Array} extent The extent to use for the gridSet
+         * @return {Array} The array with two elements representing the
+         *     tileOrigin coordinate.
+         */
+        getTileOrigin: function(tileSize, maxResolution, extent) {
+            if (Ext.isEmpty(tileSize) || Ext.isEmpty(maxResolution) ||
```
marcjansen: The tilesize should fallback to `[256, 256]`

src/util/Caching.js
```diff
@@ -0,0 +1,129 @@
...
+        },
+
+        /**
+         * Computes the tile origin (top left) for a given extent
+         * (lower left/upper right bbox), maxResolution and tileSize that
+         * can be set as the origin on a ol.tilegrid.TileGrid instance to
+         * match a GWC gridset configured in a GeoServer. GetMap requests
+         * should use the parameter tiled=true to make use of the cache.
+         *
+         * @param {Number} tileSize The tile size in pixels
```
marcjansen: One does not need to have rectangular tilesizes, Please change to `Array<Number>` with `[256, 256]`

src/util/Caching.js
```diff
@@ -0,0 +1,129 @@
...
+    statics: {
+        /**
+         * Call this function from the console (in one of your applications)
+         * and it returns config information that may be used to quickly setup a
+         * matching GWC gridset in the GeoServer.
+         *
+         * Note: If the extent is not passed to this function, the ol.view of
+         * the ol.map should have explicitly set an 'extent' property!
+         *
+         * @param {Number} tileSize The tile size in pixels
```
marcjansen: One does not need to have rectangular tilesizes, Please change to `Array<Number>` with `[256, 256]`

src/util/Caching.js
```diff
@@ -0,0 +1,129 @@
...
+         * @param {Array} extent Optional: The extent to use for the gridSet
```
marcjansen: Can you hint the order of the numbers: [b, l, r, t]? So it is clear what one has to give?

src/util/Caching.js
marcjansen: I think the docs are OK they are right now.

marcjansen:
Hey @buehner, I have made some changes to this, and would ask you to test them in your project, if possible. The test data you send me, passes and the API of the method you needed most has not changed.

Can you please take a look?

If you approve the current state, feel free to merge this.

marcjansen:
Wait a second, I want to get rid of the package-lock.json…

marcjansen:
Ok, you can go ahead with a review, @buehner. TIA!

src/util/Caching.js
```diff
@@ -82,48 +81,193 @@ Ext.define('BasiGX.util.Caching', {
...
+                    ' defaulting to [256, 256] now');
+                tileSize = [256, 256];
+            }
+            return tileSize;
+        },
 
-            return [minX, tileOriginY];
+        /**
+         * Sanitizes the passed extent, potentially taking into account the
+         * passed `ol.View`. A valid extent (left, bottom, right, top) will
```
buehner: `(left, bottom, right, top)` is a bit confusing. something like `(minX, minY, maxX, maxY)` would be more clear.

buehner:
@marcjansen I found a bug and fixed it in 19e7890. I also added a test for this scenario. Please review my change and merge if you are happy. Thx

marcjansen:
Thanks @buehner

PR #216, opened 2017-09-05
Translate scale bar & north arrow

hwbllmnn:

### DIFF
--- a/src/view/form/Print.js
+++ b/src/view/form/Print.js
@@ -744,8 +744,10 @@ Ext.define('BasiGX.view.form.Print', {
             xtype: 'checkbox',
             name: attributeRec.get('name'),
             checked: true,
-            fieldLabel: attributeRec.get('name'),
-            boxLabel: '…verwenden?' // TODO i18n
+            bind: {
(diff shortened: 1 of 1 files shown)


src/view/form/Print.js
```diff
@@ -744,8 +744,10 @@ Ext.define('BasiGX.view.form.Print', {
             xtype: 'checkbox',
             name: attributeRec.get('name'),
             checked: true,
-            fieldLabel: attributeRec.get('name'),
-            boxLabel: '…verwenden?' // TODO i18n
+            bind: {
+                fieldLabel: '{' + attributeRec.get('name') + '}',
+                boxLabel: '{labelUse}'
```
weskamm: where does `labelUse` come from?

hwbllmnn: In my case that comes from the project specific i18n files. Does it need to be added somewhere in BasiGX, too?

weskamm: Yes, please add it to the viewmodels data ontop of your class

weskamm:
nice, please merge once travis goes green

PR #226, opened 2017-09-20
Fix several scale combo issues

marcjansen:
This includes:

* Remove unneeded quotes around object keys
* Turn inline anonymous functions into class methods
* adding and using variables for often used expressions
* Properly unregistering of listeners on OpenLayers objects when the scale combo is destroyed 
* Not updating the scale combo while we are animating, but at the end of the animation as soon as possible (`Ext.Function.createBuffered`)

Please review.

### DIFF
--- a/src/view/combo/ScaleCombo.js
+++ b/src/view/combo/ScaleCombo.js
@@ -81,6 +81,14 @@ Ext.define('BasiGX.view.combo.ScaleCombo', {
     fields: ['scale', 'resolution'],
 
     /**
+     * Will hold the event keys of any listeners we bind to openlayers objects,
+     * so that we can unbind them once we get destroyed.
+     *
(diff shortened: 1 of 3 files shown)


annarieger:
LGTM, thanks for fixing this! Please merge :pineapple:

marcjansen:
@annarieger I had to change this quite substantially, as we are now needing the newest OpenLayers, so I think this deserves another look.

It would be extra nice if @annarieger, @KaiVolland and @weskamm could look into this from their projects point of views.

The method `view.getInteracting()` only exists in newer OpenLayers versions

marcjansen:
Anybody who wants to take another look? @terrestris/devs

marcjansen:
Thanks for the review.

PR #234, opened 2017-10-04
Use https where appropriate

marcjansen:
This changes 4 URLs (twice nominatim and twice the terrestris WMS) to use `https` instead of `http`.

Please review.


marcjansen:
Cheers for the review.

PR #235, opened 2017-10-05
Constrain windows we create

marcjansen:
This PR suggests to use `constrain: true` when we create windows.

Please review.


marcjansen:
Cheers for the review.

PR #239, opened 2017-10-30
Added skip map mode to print w/o extent selection & choosing template

hwbllmnn:
This can be used for custom print requirements such as printing a 'normal' pdf without a map and a given template. It skips the app selection and disables the extent rectangle if the config option `skipMapMode` is set.

### DIFF
--- a/src/view/form/Print.js
+++ b/src/view/form/Print.js
@@ -78,6 +78,11 @@ Ext.define('BasiGX.view.form.Print', {
         printExtentAlwaysCentered: true,
         printExtentMovable: false,
         printExtentScalable: false,
+        /**
+         * Option to be able to print without a map.
+         * @type {Boolean} if true, app selection and the extent rectangle are
+         * disabled.
+         */
         skipMapMode: false
     },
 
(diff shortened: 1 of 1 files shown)


src/view/form/Print.js
```diff
@@ -335,7 +339,10 @@ Ext.define('BasiGX.view.form.Print', {
         Ext.each(fieldsets, function(fs) {
             var name = fs.name;
             // TODO double check when rotated
-            var featureBbox = fs.extentFeature.getGeometry().getExtent();
+            var featureBbox;
+            if (fs.extentFeature) {
+                fs.extentFeature.getGeometry().getExtent();
```
weskamm: is this missing a variable assignment?

weskamm:
One comment added. Maybe you can add a short explanation in the code what `skipMapMode` does. Else good to go

src/view/form/Print.js
hwbllmnn: Nice catch, fixed!

hwbllmnn:
Done, description added.

PR #245, opened 2017-11-27
Merge layers

hwbllmnn:
This adds a button and some utils to merge the features of one layer into another. After clicking the button the user can map attributes of the source layer to attributes of the target layer and select which attributes to add to the target layer's schema.

### DIFF
--- a/src/util/Merge.js
+++ b/src/util/Merge.js
@@ -1,4 +1,4 @@
-/* Copyright (c) 2016-present terrestris GmbH & Co. KG
+/* Copyright (c) 2017-present terrestris GmbH & Co. KG
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -15,9 +15,8 @@
  */
 /**
  *
- * Merge Util
+ * Utility functions for merging layers and features.
  *
- * Message box utilities for a common user experience.
  *
  * @class BasiGX.util.Merge
  */
@@ -28,11 +27,16 @@ Ext.define('BasiGX.util.Merge', {
 
         /**
          * Extracts the attribute schema from the given layer. Looks for
-         * attributes on all features. Excludes fields with name 'id'.
+         * attributes on all features. Excludes fields with name 'id' by
+         * default.
          * @param  {ol.layer.Layer} layer the vector layer with the features
+         * @param  {Array} ignoreFields an array with field names to ignore
          * @return {Array}       a sorted list of attribute names
          */
-        extractSchema: function(layer) {
+        extractSchema: function(layer, ignoreFields) {
+            if (!ignoreFields) {
+                ignoreFields = ['id'];
+            }
             var features = layer.getSource().getFeatures();
             var schema = [];
             Ext.each(features, function(feature) {
@@ -40,7 +44,7 @@ Ext.define('BasiGX.util.Merge', {
                     if (value === undefined || value && value.getExtent) {
                         return;
                     }
-                    if (!schema.includes(key) && key !== 'id') {
+                    if (!schema.includes(key) && !ignoreFields.includes(key)) {
                         schema.push(key);
                     }
                 });
@@ -75,7 +79,7 @@ Ext.define('BasiGX.util.Merge', {
             var selected = multiselect.getSelected();
             var list = [];
             Ext.each(selected, function(item) {
-                list.push(item.data.field1);
+                list.push(item.get('field1'));
             });
             return list;
         },
--- a/src/util/Merge.js
+++ b/src/util/Merge.js
@@ -26,14 +26,6 @@ Ext.define('BasiGX.util.Merge', {
     ],
     statics: {
 
-        /*begin i18n*/
-        adoptAttributes: 'Select attributes to add to the schema:',
-        applyText: 'Apply',
-        cancelText: 'Cancel',
-        windowTitle: 'Merge features',
-        matchText: 'Please select attributes to be mapped:',
-        /*end i18n*/
-
         /**
          * Extracts the attribute schema from the given layer. Looks for
          * attributes on all features. Excludes fields with name 'id'.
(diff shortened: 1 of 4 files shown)


src/util/Merge.js
```diff
@@ -0,0 +1,288 @@
...
+         * Creates and shows the merge window.
+         * @param  {Array} labels       the original attributes label list
+         * @param  {Array} dropdowns    the dropdowns with the source layer
+         * attributes
+         * @param  {ol.layer.Vector} sourceLayer  the layer to merge
+         * @param  {ol.layer.Vector} targetLayer  the layer to merge to
+         * @param  {Array} sourceSchema the source layer schema
+         * @param  {Array} origSchema   the target layer schema
+         */
+        createMergeWindow: function(
```
KaiVolland: I'd prefer this to be a real component not a function in an util method. You could then put the handlers in the component, too.

src/util/Merge.js
```diff
@@ -0,0 +1,109 @@
+/* Copyright (c) 2016-present terrestris GmbH & Co. KG
```
marcjansen: 2017...

src/util/Merge.js
```diff
@@ -0,0 +1,109 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+/**
+ *
+ * Merge Util
+ *
+ * Message box utilities for a common user experience.
```
marcjansen: comment is wrong, copy and paste?

src/util/Merge.js
```diff
@@ -0,0 +1,109 @@
...
+    ],
+    statics: {
+
+        /**
+         * Extracts the attribute schema from the given layer. Looks for
+         * attributes on all features. Excludes fields with name 'id'.
+         * @param  {ol.layer.Layer} layer the vector layer with the features
+         * @return {Array}       a sorted list of attribute names
+         */
+        extractSchema: function(layer) {
```
marcjansen: can we have an ignorefields array which defaults to `['id']`?

src/util/Merge.js
```diff
@@ -0,0 +1,109 @@
...
+         * Extracts the list of attributes to add from the merge window.
+         * @param  {Ext.window.Window} win the merge windowTitle
+         * @return {Array}     a list of source layer attributes
+         */
+        extractToCopyAttributes: function(win) {
+            var multiselect = win.down('multiselect');
+            var selected = multiselect.getSelected();
+            var list = [];
+            Ext.each(selected, function(item) {
+                list.push(item.data.field1);
```
marcjansen: or better `item.get('field1')`?

src/view/button/MergeSelection.js
```diff
@@ -0,0 +1,78 @@
+/* Copyright (c) 2015-present terrestris GmbH & Co. KG
```
marcjansen: 2017...

src/view/button/MergeSelection.js
```diff
@@ -0,0 +1,78 @@
...
+            tooltip: 'Selektierte Features in diesen Layer übernehmen',
+            text: 'Features übernehmen'
+        }
+    },
+
+    /**
+     *
+     */
+    bind: {
+        text: '{text}'
```
marcjansen: can we bind the tooltip? will this work in modern variants?

src/view/button/MergeSelection.js
```diff
@@ -0,0 +1,78 @@
...
+ * Merge selected features into the currently selected layer.
+ *
+ * @class BasiGX.view.button.MergeSelection
+ */
+Ext.define('BasiGX.view.button.MergeSelection', {
+    extend: 'Ext.button.Button',
+    xtype: 'basigx-button-mergeselection',
+
+    requires: [
+        'Ext.window.Window',
```
marcjansen: will this lead to problens when used in a modern environment?

src/view/window/MergeWindow.js
```diff
@@ -0,0 +1,245 @@
...
+        this.mergeLayers();
+    },
+
+    /**
+     * Extracts the attribute schema from the given layer. Looks for
+     * attributes on all features. Excludes fields with name 'id'.
+     * @param  {ol.layer.Layer} layer the vector layer with the features
+     * @return {Array}       a sorted list of attribute names
+     */
+    extractSchema: function(layer) {
```
marcjansen: Is this the same mthod as above?

src/view/window/MergeWindow.js
```diff
@@ -0,0 +1,245 @@
...
+
+            win.destroy();
+        };
+    },
+
+    /**
+     * Callback that just closes/destroys the merge window.
+     */
+    cancelHandler: function() {
+        this.up('window').destroy();
```
marcjansen: probably better to call close() people can define the close action

src/util/Merge.js
hwbllmnn: Done.

src/util/Merge.js
hwbllmnn: Fixed.

src/util/Merge.js
hwbllmnn: Done.

src/util/Merge.js
hwbllmnn: Changed.

src/view/button/MergeSelection.js
hwbllmnn: Back to the present.

src/view/button/MergeSelection.js
hwbllmnn: Bound. I don't think it will work in the modern variant, but then I can't really imagine using the merge tool on mobile. If needed, this will be a lot more work anyway (but it isn't a requirement right now).

src/view/window/MergeWindow.js
hwbllmnn: True, after refactoring this is not used any more anyway, I removed it.

src/view/window/MergeWindow.js
hwbllmnn: Done.

src/view/button/MergeSelection.js
hwbllmnn: See my comment below.

PR #250, opened 2017-11-29
Minor layout enhancements for MergeWindow

weskamm:
title says it all

### DIFF
--- a/src/view/window/MergeWindow.js
+++ b/src/view/window/MergeWindow.js
@@ -42,6 +42,7 @@ Ext.define('BasiGX.view.window.MergeWindow', {
     layout: 'fit',
     autoShow: true,
     items: [],
+    bodyPadding: 5,
 
     config: {
         /**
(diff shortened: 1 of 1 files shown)


hwbllmnn:
Merging now, nice.

src/view/window/MergeWindow.js
```diff
@@ -57,8 +58,25 @@ Ext.define('BasiGX.view.window.MergeWindow', {
...
-        this.callParent();
-        this.mergeLayers();
+        var me = this;
+        me.callParent();
+        me.mergeLayers();
+        me.registerMultiSelectFilter();
+        var combos = me.query('combo');
+        Ext.each(combos, function(combo) {
+            combo.on('change', function() {
+                me.registerMultiSelectFilter();
```
marcjansen: Will this ptobably be registered too often?

hwbllmnn: This is actually a misnomer, left over from an experiment while tracking down a bug. I'll rename it back for clarity in another PR.

PR #253, opened 2017-11-29
Fix empty grid

hwbllmnn:
If starting out with an empty layer, use the first added feature to
determine the attribute schema.

@weskamm please review

### DIFF
--- a/src/view/grid/FeatureGrid.js
+++ b/src/view/grid/FeatureGrid.js
@@ -195,6 +195,13 @@ Ext.define('BasiGX.view.grid.FeatureGrid', {
         grid.on('itemmouseleave', function() {
             me.highlightSource.clear();
         });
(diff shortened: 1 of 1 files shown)


src/view/grid/FeatureGrid.js
```diff
@@ -195,6 +195,13 @@ Ext.define('BasiGX.view.grid.FeatureGrid', {
         grid.on('itemmouseleave', function() {
             me.highlightSource.clear();
         });
+        if (this.getLayer().getSource().getFeatures().length === 0) {
+            this.addFeatureKey = this.getLayer().getSource().once('addfeature',
+            function() {
+                me.reconfigureStore(grid.getStore());
+                me.getLayer().getSource().unByKey(me.addFeatureKey);
```
marcjansen: If you use `once` there is no need to manually unbind.

PR #255, opened 2017-11-30
Added selectionLayer config

hwbllmnn:
This allows to select/deselect features in the grid if they are
added/removed to/from the selection layer's source, also the other way round.

### DIFF
--- a/src/view/grid/FeatureGrid.js
+++ b/src/view/grid/FeatureGrid.js
@@ -80,6 +80,8 @@ Ext.define('BasiGX.view.grid.FeatureGrid', {
         this.registerEvents();
         this.createHighlightLayer(this.getMap());
         this.appendMenuEntries();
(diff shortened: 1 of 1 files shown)


src/view/grid/FeatureGrid.js
```diff
@@ -60,6 +80,60 @@ Ext.define('BasiGX.view.grid.FeatureGrid', {
...
+            selLayer.getSource().un(
+                'addfeature',
+                this.selectionFeatureAdded,
+                this
+            );
+            selLayer.getSource().un(
+                'removefeature',
+                this.selectionFeatureRemoved,
+                this
+            );
```
marcjansen: if this was a utility method unbind sellayer events, we could call it here and above.

hwbllmnn: Factored out in 5fe50c8, thanks

PR #260, opened 2017-12-22
Fix value checks in the context of the SLD styler

buehner:
This possibly (!) fixes some problems that occured when using the SLD styler classes (but may also introduce a bug in other projects that use the SLD styler when their code is not adapted!?)

`DEFAULT_STROKE_WIDTH` should have a default of `1` in my eyes as this is the default the GeoServer uses. Unfortunately there is some strange magic happening on the GeoServer side when updating a style with such a default value: GeoServer seems to remove default values like `<sld:CssParameter name="stroke-width">1</sld:CssParameter>` from the SLD. The next time the style is loaded in the SLD styler, the value from DEFAULT_STROKE_WIDTH would be used, which was wrong before this PR...

This also adds some defaults to the stroke width numberfield to support decimals.

Furthermore some conditional checks in `SLDStyler.js` have been enhanced. There were edge-cases, e.g. when a value has the value `0` where the previous logic did not work as expected.

### DIFF
--- a/src/util/SLD.js
+++ b/src/util/SLD.js
@@ -41,7 +41,7 @@ Ext.define('BasiGX.util.SLD', {
         ],
         DEFAULT_STROKE_OPACITY: '1',
         DEFAULT_STROKE_COLOR: '#000000',
-        DEFAULT_STROKE_WIDTH: '0',
+        DEFAULT_STROKE_WIDTH: '1',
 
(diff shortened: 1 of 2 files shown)


buehner:
We should talk about this before merging as this may change the behaviour of the SLD styler in other projects. I think we should to check this before.

PR #267, opened 2018-01-15
Refactor redlining container

weskamm:
This moves the single buttons like modify or delete object and their logic to their own classes.

### DIFF
--- a/src/view/button/DigitizeDeleteObject.js
+++ b/src/view/button/DigitizeDeleteObject.js
@@ -1,0 +1,136 @@
+/* Copyright (c) 2018-present terrestris GmbH & Co. KG
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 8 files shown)


src/view/button/DigitizePostit.js
```diff
@@ -0,0 +1,275 @@
...
+     * replaced with the passed `spaceReplacer`.
+     *
+     * See http://stackoverflow.com/questions/14484787/wrap-text-in-javascript
+     *
+     * @param {String} str The string to wrap.
+     * @param {Number} width The width of a line (number of characters).
+     * @param {String} spaceReplacer The string to replace spaces with.
+     * @return {String} The 'wrapped' string.
+     */
+    stringDivider: function(str, width, spaceReplacer) {
```
hwbllmnn: The same function is used elsewhere, would be nice if it was factored out into a util.

src/view/button/DigitizeDeleteObject.js
```diff
@@ -0,0 +1,136 @@
...
+                me.deleteSnapInteraction.setActive(false);
+            }
+        }
+    },
+
+    /**
+     * @param {ol.Feature} clone The cloned feature to get the feature from.
+     * @return {ol.Feature} The final feature derived from the `clone`.
+     */
+    getFeatureFromClone: function(clone) {
```
hwbllmnn: The same function is used elsewhere, wouid be nice it if was factored out in a util.

weskamm:
issues resolved, merging now

PR #268, opened 2018-01-16
Fixup redlining tools

weskamm:
Some enhancements...


weskamm:
That util has just been added before and is not used currently in any project / application yet. So yess its a bug fix and no, currently no tests

PR #271, opened 2018-01-18
Fixed randomly failing tests

hwbllmnn:

### DIFF
--- a/test/.eslintrc
+++ b/test/.eslintrc
@@ -2,6 +2,7 @@
   "extends":"openlayers",
   "globals":{
     "afterEach":false,
+    "after":false,
     "beforeEach":false,
     "describe":false,
     "document":false,
--- a/test/spec/util/Application.test.js
+++ b/test/spec/util/Application.test.js
@@ -1,22 +1,15 @@
 Ext.Loader.syncRequire(['BasiGX.util.Application']);
 
 describe('BasiGX.util.Application', function() {
+    var testObjs = TestUtil.setupTestObjects();
+    var mapComponent = testObjs.mapComponent;
+
     describe('Basics', function() {
         it('is defined', function() {
             expect(BasiGX.util.Application).to.not.be(undefined);
         });
     });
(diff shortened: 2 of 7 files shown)


marcjansen:
I am `+0`, e.g. not blocking this.

Personally I would have tried another approach; instead of no longer cleaning up behind (some) individual tests, I would have searched where something isn't properly cleaned up right now. I assume the random test failures stem from not properly cleaning up. But again, I might miss something or misunderstand this. I would love to have another opinion on this.

weskamm:
I see other tests failing on my local machine, we should wait with merging and analyse this in more detail

hwbllmnn:
As far as I could tell, the tests failed _because_ of the cleaning up. I'm not sure why, but I imagine it being a result of either ExtJS failing to properly do so (.destroy), as @weskamm suggested, or PhantomJS doing garbage collection at inconvenient times, or even a combination of both.

Anyway, please note that I only have anecdotal evidence of this fixing the problem (I ran the tests locally ~50 times with none failing). If that's you restarting travis all the time currently, please keep clicking, the issue happens on slow machines much more often than on fast ones :-)

weskamm:
Yes its me restarting travis ;-) But now its time for fireevening

hwbllmnn:
@weskamm agreed, I can very well imagine that other tests may be involved. I'll prepare a longer run right now.

hwbllmnn:
Have a good one ;-)

hwbllmnn:
I pushed another few fixes, that should fix the remaining ones. I just ran 1000 test runs without any failures.

weskamm:
Ok, will merge this now. If the problem occurs again, we should have a deeper look in `setupTestObjects` and `teardownTestObjects` of the `TestUtil`. I had some success temporarily disable the Redlining test, but i think the problem may be found somewhere in updates of phantomjs and asynchronous behaviour of some cleanup /garbage collection things...

PR #273, opened 2018-01-18
Fix the highlighting in FeatureGrid

weskamm:
This resolves the issue that identical features, that are already contained in the selectionlayer, will get added multiple times


hwbllmnn:
🔝 Please merge once CI is happy again

PR #279, opened 2018-01-26
WIP: Fix feature equality check

hwbllmnn:
Due to the checking for equal geometries many of the digitizing tools run into problems. This is an attempt to fix it.

* if both the clone and the original have an id set, it is used for comparison (thus ruling out false positives)


weskamm:
Looks good, will merge

PR #281, opened 2018-01-29
Allow digitizing of multi geometries via config option

hwbllmnn:


weskamm:
nice change, merge at will

PR #292, opened 2018-03-01
Fix unByKey call to use static function

hwbllmnn:


hwbllmnn:
Oh, I didn't realize that. I've looked back until 3.19 where it already existed.

marcjansen:
just merge... it'll be ok

PR #307, opened 2018-03-29
Added download util

hwbllmnn:
Adds a download util to download vector layers as .shp or .geojson clientside only.

### DIFF
--- a/src/util/Download.js
+++ b/src/util/Download.js
@@ -30,11 +30,11 @@ Ext.define('BasiGX.util.Download', {
          */
         downloadLayer: function(layer, map, format) {
             var name = layer.get('name');
-            var json = this.layerToGeoJson(layer, map);
+            var json = BasiGX.util.Download.layerToGeoJson(layer, map);
             var result;
             switch (format) {
                 case 'zip': {
-                    result = this.geoJsonToShpBytes(json, name);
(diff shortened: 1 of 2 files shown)


src/util/Download.js
```diff
@@ -0,0 +1,98 @@
...
+         * Download a vector layer from memory, in EPSG:4326. For IE you'll need
+         * a polyfill from https://developer.mozilla.org/en-US/docs
+         * /Web/JavaScript/Reference/Global_Objects/TypedArray/from
+         * @param  {ol.layer.Vector} layer  the layer to download
+         * @param  {ol.Map} map    the map the layer is contained in
+         * @param  {String} format geojson or zip (shapefile)
+         */
+        downloadLayer: function(layer, map, format) {
+            var name = layer.get('name');
+            var json = this.layerToGeoJson(layer, map);
```
weskamm: Please use the explicit method name `BasiGX.util.Download.layerToGeoJson` or we might get scoping problems

src/util/Download.js
```diff
@@ -0,0 +1,98 @@
...
+            var result;
+            switch (format) {
+                case 'zip': {
+                    result = this.geoJsonToShpBytes(json, name);
```
weskamm: see above

weskamm:
Just a minor, else good to go

PR #308, opened 2018-03-29
Enable rotation in printform

weskamm:
This enables the print extent to be rotatable and modifies the print request to correctly handle rotation

### DIFF
--- a/src/ol3/extension/TransformInteraction.js
+++ b/src/ol3/extension/TransformInteraction.js
@@ -692,9 +692,13 @@ Ext.define('BasiGX.ol3.extension.TransformInteraction', {
      * @param {ol.MapBrowserEvent} evt Map browser event.
      * @return {boolean} `false` to stop the drag sequence.
      */
-    ol.interaction.Transform.prototype.handleUpEvent_ = function() {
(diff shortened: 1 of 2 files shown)


src/ol3/extension/TransformInteraction.js
```diff
@@ -692,9 +692,13 @@ Ext.define('BasiGX.ol3.extension.TransformInteraction', {
      * @param {ol.MapBrowserEvent} evt Map browser event.
      * @return {boolean} `false` to stop the drag sequence.
      */
-    ol.interaction.Transform.prototype.handleUpEvent_ = function() {
+    ol.interaction.Transform.prototype.handleUpEvent_ = function(evt) {
+        var deltaX = this.center_[0] - evt.coordinate[0];
+        var deltaY = this.center_[1] - evt.coordinate[1];
+        var a = Math.atan2(deltaY, deltaX);
```
marcjansen: Is `Math.atan2` everywhere available? It should be easily polyfilled otherwise.

weskamm: Looks ok according to https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Math/atan2

weskamm:
Thanks for review, merging..

PR #313, opened 2018-04-25
Render the styler preview as real ol-map

weskamm:
This replaces the GetMap / GetLegendGraphic preview in the SLDStyler with a real openlayers map for better preview purposes.

styler

### DIFF
--- a/src/view/container/SLDStyler.js
+++ b/src/view/container/SLDStyler.js
@@ -56,6 +56,7 @@ Ext.define('BasiGX.view.container.SLDStyler', {
 
     requires: [
         'Ext.ux.colorpick.Button',
+        'GeoExt.component.Map',
         'BasiGX.view.panel.GraphicPool',
         'BasiGX.view.panel.FontSymbolPool',
         'BasiGX.util.Color',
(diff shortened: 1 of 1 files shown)


src/view/container/SLDStyler.js
```diff
@@ -1249,17 +1265,70 @@ Ext.define('BasiGX.view.container.SLDStyler', {
...
+                        Ext.Ajax.request({
+                            binary: true,
+                            url: baseUrl,
+                            method: 'POST',
+                            params: paramObj,
+                            defaultHeaders: BasiGX.util.CSRF.getHeader(),
+                            scope: this,
+                            success: function(response) {
+                                if (response.responseBytes) {
+                                    var blob = new Blob(
```
marcjansen: Do we need a polyfill, or is this supported out f the box everywhere where BasiGX runs?

weskamm: I think its ok, as i fallback to GET Requests in non capable browsers, which seems to mainly hit IE < 11. Blob construction has been already happening in this class for a while

PR #318, opened 2018-05-03
Add the full split parameter to the intersection op

hwbllmnn:
This adds two new parameters to the intersection operation:

`tolerance`: used to filter out resulting polygons smaller that this
and to compare polygons for equality.

`fullSplit`: used to obtain a list of all geometry parts resulting
after doing an intersection between possibly multiple polygons.

@weskamm @terrestris/devs Please review.


weskamm:
nice enhancement, please merge at will!

PR #319, opened 2018-05-03
Get smarter when parsing cql filters using regexp

hwbllmnn:
Title says it all.

@weskamm @terrestris/devs please review

### DIFF
--- a/src/util/WFS.js
+++ b/src/util/WFS.js
@@ -130,7 +130,7 @@ Ext.define('BasiGX.util.WFS', {
             if (!cqlFilter) {
                 return;
             }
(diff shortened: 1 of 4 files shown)


src/util/WFS.js
```diff
@@ -130,16 +130,17 @@ Ext.define('BasiGX.util.WFS', {
             if (!cqlFilter) {
                 return;
             }
-            var parts = cqlFilter.split(' ');// TODO: get smarter
-            if (parts.length !== 3) {
+            var parts = /([\w_]+)\s*([<>=!]+|in|like|IN|LIKE)\s*(.+)/g
+                .exec(cqlFilter);
```
marcjansen: How about the `LiKe` with the case insensitive flag?

hwbllmnn:
True. Done :sunglasses:

PR #322, opened 2018-05-16
Use transparent: true param when adding wms

hwbllmnn:
Title says it all.


weskamm:
nice, merge at will!

marcjansen:
Can we somehow detect this from the capabilities?

marcjansen:
makes sense. Please merge

PR #328, opened 2018-05-28
Fix failing build

dnlkoch:
Replaces outdated links for `ogc-schema`/`w3c-schema` sources.

Please review @terrestris/devs.


weskamm:
nice catch :-)

dnlkoch:
Thanks for the review @weskamm!

PR #344, opened 2018-10-25
Support embedding of feature grid functionality in a panel

hwbllmnn:
Previously the merge button would fail if not embedded in a `window`.

### DIFF
--- a/src/view/button/MergeSelection.js
+++ b/src/view/button/MergeSelection.js
@@ -55,14 +55,16 @@ Ext.define('BasiGX.view.button.MergeSelection', {
      */
     config: {
         handler: function() {
+            var grid;
             var parent = this.up('window');
             // support embedding in window or panel
             // This assumes the button is embedded in a buttongroup or similar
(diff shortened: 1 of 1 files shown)


src/view/button/MergeSelection.js
```diff
@@ -55,7 +55,14 @@ Ext.define('BasiGX.view.button.MergeSelection', {
      */
     config: {
         handler: function() {
-            var grid = this.up('window').down('basigx-grid-featuregrid');
+            var parent = this.up('window');
+            // support embedding in window or panel
+            // This assumes the button is embedded in a buttongroup or similar
+            // panel, hence walking up two panels.
+            if (!parent) {
+                parent = this.up('panel').up('panel');
```
KaiVolland: This could be more precise but if it works...

hwbllmnn: You're right, I've changed this to do it via a new config `featureGridSelectorFn` to be more flexible.

PR #346, opened 2018-11-09
Support multiple merge target layers

hwbllmnn:
Also sets an ID on cloned/merged features, else selection will not function properly on the grid nor in the map. Also adds a new config option `mergedFeaturesFn` to get notified once merge has commenced.

### DIFF
--- a/src/view/button/MergeSelection.js
+++ b/src/view/button/MergeSelection.js
@@ -55,6 +55,7 @@ Ext.define('BasiGX.view.button.MergeSelection', {
      */
     config: {
         handler: function() {
+            var me = this;
             var grid;
             var parent = this.up('window');
             // support embedding the feature grid anywhere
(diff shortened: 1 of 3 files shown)


src/view/button/MergeSelection.js
```diff
@@ -70,14 +69,33 @@ Ext.define('BasiGX.view.button.MergeSelection', {
             Ext.create({
                 xtype: 'basigx-window-merge',
                 sourceLayer: this.getSourceLayer(),
-                targetLayer: targetLayer
+                targetLayer: targetLayer,
+                extraTargetLayers: this.getExtraTargetLayers(),
+                mergedFeaturesFn: function(features) {
+                    me.getMergedFeaturesFn()(features);
```
weskamm: if `getMergedFeaturesFn` is null, which seems to be the default, will this call break?

weskamm:
just added a minor question

src/view/button/MergeSelection.js
hwbllmnn: Indeed it will, thanks! Fixed in 37727bd

PR #352, opened 2018-11-16
Do not update coord textfields while they have focus

marcjansen:


marcjansen:
Thanks for the review!

PR #353, opened 2018-11-16
Add a history navigation button

chrismayer:
This adds a simple history navigation button in order to interact with the HTML5 History API.

Since the History API is supported by current browsers down to IE10 the usage of this might need a polyfill (like https://github.com/browserstate/history.js/) in older browsers, see  https://caniuse.com/#search=history for details.

@terrestris/devs how do you deal with that?

### DIFF
--- a/src/view/button/History.js
+++ b/src/view/button/History.js
@@ -1,4 +1,4 @@
-/* Copyright (c) 2015-present terrestris GmbH & Co. KG
+/* Copyright (c) 2018-present terrestris GmbH & Co. KG
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


src/view/button/History.js
```diff
@@ -0,0 +1,89 @@
...
+    /**
+     * Called when the button is clicked.
+     * Performs the navigation action
+     *
+     * @param {Ext.Button} btn The history button itself
+     */
+    onBtnClick: function(btn) {
+        var me = this;
+        if (me.mode === 'HISTORY') {
+            if (btn.direction === 'back') {
```
marcjansen: Should this be uppercase? and below as well?

src/view/button/History.js
```diff
@@ -0,0 +1,89 @@
+/* Copyright (c) 2015-present terrestris GmbH & Co. KG
```
ahennr: Change to 2018 plz

chrismayer:
Thanks for the review guys :+1: 

I put a comment about the possibly needed polyfill in the API-Docs of the class. So this should be good to go once travis gets green.

src/view/button/History.js
chrismayer: DOne in f122b4a

src/view/button/History.js
chrismayer: Sure, was a long day today :wink:

PR #354, opened 2018-11-16
Add getLayersBy util

ahennr:
Title says it all.

### DIFF
--- a/src/util/Layer.js
+++ b/src/util/Layer.js
@@ -107,7 +107,6 @@ Ext.define('BasiGX.util.Layer', {
                 if (layer.get(key) === value &&
                     layer instanceof ol.layer.Base) {
                     matchingLayers.push(layer);
-                    return false;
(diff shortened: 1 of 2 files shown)


src/util/Layer.js
```diff
@@ -78,6 +78,44 @@ Ext.define('BasiGX.util.Layer', {
...
+            } else {
+                var map = BasiGX.util.Map.getMapComponent().getMap();
+                layers = map.getLayers().getArray();
+            }
+
+            Ext.each(layers, function (layer) {
+                if (layer.get(key) === value &&
+                    layer instanceof ol.layer.Base) {
+                    matchingLayers.push(layer);
+                    return false;
```
marcjansen: Hmm. This will stop the itartion, is this intended?

ahennr: No, returning `false` is wrong here. Fixed in cde24dc

ahennr:
Thx for review @marcjansen

PR #355, opened 2018-11-19
Add basic unit tests for HistoryButton

chrismayer:
Follow-up for #353


chrismayer:
Thanks for the review @marcjansen.

PR #356, opened 2018-11-19
Allow a fallback for timeParam

ahennr:
This PR adds a further parameter to `getTimeFilterParts` function in WFS util to pass a time parameter name whioch is not set in layer source.

Plz review @hwbllmnn @weskamm


ahennr:
:top:

Issue #357, opened 2018-11-20
WFS Filters not working with a MapServer 7 backend


geographika:
The OGC filters currently generated by BasiGX.util.WFS cause errors when requesting features from MapServer (with a GET request) such as:

`msWFSGetFeature(): WFS server error. Invalid or Unsupported FILTER in GetFeature : <Filter><BBOX><PropertyName>msGeometry</PropertyName><Envelope><gml:lowerCorner>-1399714.8619581475 6740828.650413123</gml:lowerCorner>      <gml:upperCorner>-289237.7150311069 7405525.048381017</gml:upperCorner></gml:Envelope></ogc:BBOX><gml:Envelope xmlns:gml=http://www.opengis.net/gml srsName=EPSG:3857><gml:lowerCorner>-1399714.8619581475 6740828.650413123</gml:lowerCorner><gml:upperCorner>-289237.7150311069 7405525.048381017</gml:upperCorner><Envelope></BBOX><Filter>
`

It appears these are caused by the following:

- any namespace properties in a tag e.g. xmlns:gml=http://www.opengis.net/gml
- an additional AND wrapper around the filter
- use of BBOX and Intersects?

An example of a working BBOX filter is:
`
<Filter><BBOX><PropertyName>msGeometry</PropertyName><Box%20srsName='EPSG:3857'><coordinates>-1009073.7780155,6855880.1040363 -754844.22198452,6990867.8959637</coordinates></Box></BBOX></Filter>
`

The above problems may be a MapServer issue rather than invalid filters.

marcjansen:
I think this isn't a mapserver bug, but sth. in BasiGX

PR #358, opened 2018-11-20
[WIP] Modify Filter XML to work with MapServer WFS

geographika:
Relates to #357 
This pull request makes the filters work correctly with MapServer, but may break GeoServer. In order for cleaner code it could be better to specify a backend parameter?

### DIFF
--- a/src/util/WFS.js
+++ b/src/util/WFS.js
@@ -103,6 +103,16 @@ Ext.define('BasiGX.util.WFS', {
                 '    </gml:Envelope>' +
                 '  </ogc:Intersects>' +
                 '</ogc:And>';
+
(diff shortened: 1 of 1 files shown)


marcjansen:
Thanks for the contribution. This will need testing, maybe anyone from @terrestris/dev can check this?

geographika:
@hwbllmnn - I'll mark this pull request as Work In Progress - as you say simply overwriting the tpl string changes the workings of the function. This was mainly to show an example that was working in MapServer (getting features within the extent of a map following a pan). 
I think checking for number of filters prior to adding an AND tag is probably valid.

marcjansen:
Superseded by #372

PR #360, opened 2018-11-22
Add some utils

annarieger:
Added some further useful util methods.

In details: 

- `getAttributeFilter` method for `WFSUtil`
- `getLayerByLayersParam` for `LayerUtil`


src/util/Layer.js
```diff
@@ -78,6 +78,39 @@ Ext.define('BasiGX.util.Layer', {
             return matchingLayer;
         },
 
+                /**
```
marcjansen: indentation off

src/util/Layer.js
```diff
@@ -78,6 +78,39 @@ Ext.define('BasiGX.util.Layer', {
...
+         * @param {String} layersParam The (qualified) layers param,
+         *  e.g. 'netview:NodeB'
```
marcjansen: can we have another example, please

src/util/Layer.js
```diff
@@ -78,6 +78,39 @@ Ext.define('BasiGX.util.Layer', {
...
+         *
+         * @return {ol.layer.Layer} The ol layer from the map (if it
+         *  could be found)
+         */
+        getLayerByLayersParam: function(layersParam) {
+            var map = BasiGX.util.Map.getMapComponent().getMap();
```
marcjansen: can we make it so that we can pass this as optional parameter? I think the other methods do it like this.

src/util/WFS.js
```diff
@@ -117,6 +117,44 @@ Ext.define('BasiGX.util.WFS', {
...
+        getAttributeFilter: function(propertyNames, queryValue, combinator,
+            matchCase) {
+            var defaultCombineWith = 'Or';
+            var defaultMatchCase = false;
+
+            var combineWith = combinator || defaultCombineWith;
+            var match = matchCase || defaultMatchCase;
+
+            var tpl = '' +
+                '  <PropertyIsLike wildCard="*" singleChar="." escape="!" ' +
```
marcjansen: can we configure the namespace of these tags? Purely optional of course.

src/util/WFS.js
```diff
@@ -117,6 +117,44 @@ Ext.define('BasiGX.util.WFS', {
...
             return filter;
         },
+        /**
+         * Returns a attribute LIKE filter for the passed property names. Can be
+         * combined with `Or` (default) or `And` operator.
+         *
+         * @param {Array} propertyNames The name of the properties which should
+         * be queried.
+         * @param {String} queryValue The value to query.
+         * @param {Array} combinator Should be either `And` or `Or`.
```
marcjansen: wrong type?

src/util/WFS.js
```diff
@@ -117,6 +117,44 @@ Ext.define('BasiGX.util.WFS', {
...
+            var match = matchCase || defaultMatchCase;
+
+            var tpl = '' +
+                '  <PropertyIsLike wildCard="*" singleChar="." escape="!" ' +
+                '  matchCase="' + match + '">' +
+                '      <PropertyName>{0}</PropertyName>' +
+                '      <Literal>*{1}*</Literal>' +
+                '  </PropertyIsLike>';
+
+            var filter = '<ogc:' + combineWith + '>';
```
marcjansen: Same comment wrt to the xml namespace alias.

src/util/Layer.js
annarieger: fixed in d20fd30

src/util/Layer.js
annarieger: fixed in d20fd30

src/util/Layer.js
annarieger: added in d20fd30

src/util/WFS.js
annarieger: yepp, fixed in 7fca64e

src/util/WFS.js
annarieger: changed in 7fca64e

PR #366, opened 2018-11-28
Switch to karma && chrome-headless to run tests

hwbllmnn:
This also fixes the security vulnerability warnings regarding the dependencies.

From my local coverage report it looks like the coverage is counted somewhat differently when using karma, so it drops to ~19%.


package.json
```diff
@@ -53,8 +52,17 @@
         "eslint": "5.9.0",
         "expect.js": "0.3.1",
         "istanbul": "0.4.5",
+        "karma": "3.1.1",
+        "karma-chrome-launcher": "2.2.0",
+        "karma-coverage": "1.1.2",
+        "karma-coverage-istanbul-reporter": "2.0.4",
+        "karma-expect": "1.1.3",
+        "karma-firefox-launcher": "1.1.0",
```
dnlkoch: According to `karma.conf.js` this launcher is not used currently.

package.json
```diff
@@ -53,8 +52,17 @@
...
+        "karma-mocha": "1.3.0",
+        "karma-mocha-reporter": "2.2.5",
+        "karma-sinon": "1.0.5",
         "mocha": "5.2.0",
-        "mocha-phantomjs": "4.1.0",
+        "puppeteer": "1.10.0",
```
dnlkoch: Just wondering, is `puppeteer` really needed/used?

karma.conf.js
```diff
@@ -0,0 +1,94 @@
...
+        'src/**/*js',
+        'test/test-helper-functions.js',
+        'test/**/*js'
+    ];
+
+    config.set({
+        basePath: '',
+
+        proxies: {
+            '/koalaProxy': proxyUrl,
```
dnlkoch: Is this proxy (under this name) needed in `BasiGX`?

karma.conf.js
```diff
@@ -0,0 +1,94 @@
...
+        'https://cdn.jsdelivr.net/gh/highsource/jsonix@2.4.1/dist/Jsonix-min.js',
+        'https://cdn.jsdelivr.net/gh/bjornharrtell/jsts@gh-pages/1.4.0/jsts.min.js',
+        'https://cdn.jsdelivr.net/gh/highsource/ogc-schemas@2.6.1/scripts/lib/SLD_1_0_0_GeoServer.js',
+        'https://cdn.jsdelivr.net/gh/highsource/ogc-schemas@2.6.1/scripts/lib/Filter_1_0_0.js',
+        'https://cdn.jsdelivr.net/gh/highsource/ogc-schemas@2.6.1/scripts/lib/GML_2_1_2.js',
+        'https://cdn.jsdelivr.net/gh/highsource/w3c-schemas@1.4.0/scripts/lib/XLink_1_0.js',
+        'https://cdn.jsdelivr.net/npm/proj4@2.5.0/dist/proj4-src.min.js',
+        'src/**/*js',
+        'test/test-helper-functions.js',
+        'test/**/*js'
```
dnlkoch: Unsure here, but will this also import the `spec.js` files? Could this explain the coverage mismatch?

package.json
hwbllmnn: True, removed.

karma.conf.js
hwbllmnn: The setup works better now, with slightly improved coverage. I'm guessing the remaining difference is due to the previously manual instrumentalization of the source files.

PR #371, opened 2018-12-06
Add missing test stubs

marcjansen:
This hopefully makes it easier for people to add tests, as they at least have a starting point

These were generated with the following script, which might have captured everything, but I have only quickly tested it.

```bash
#!/usr/bin/env bash
srcDir="./src";
testDir="./test/spec"

for fullFileName in $(find $srcDir -name '*.js'); do
    [ -e "$fullFileName" ] || continue
    echo $fullFileName

    basename=$(basename -- "$fullFileName")
    extension="${basename##*.}"
    filename="${basename%.*}"

    folder="${fullFileName/$srcDir\//}"
    folder="${folder/\/$basename/}"

    middle="${folder/\//.}"

    testFile="${fullFileName/$srcDir/$testDir}"
    testFile="${testFile/$extension/test.js}"

    className="BasiGX.$middle.$filename"

    if [ -e "$testFile" ]; then
        continue
    else
        echo "MISSING"

        CONTENT=$(cat <<'END_HEREDOC'
Ext.Loader.syncRequire(['__CLASSNAME__']);

describe('__CLASSNAME__', function() {
    describe('Basics', function() {
        it('is defined', function() {
            expect(__CLASSNAME__).to.not.be(undefined);
        });
    });
});

END_HEREDOC
)

        echo "$CONTENT" | sed "s|__CLASSNAME__|$className|g" > $testFile
    fi

    echo
done
```

Please review.


marcjansen:
Thanks for the review!

PR #373, opened 2018-12-08
Add several tests for the WFS util

marcjansen:


marcjansen:
Thanks for the review

PR #375, opened 2018-12-10
Removes togglegroup from spatial operator buttons

hwbllmnn:
Typically we will have a `toggleGroup` eg. `draw` making sure only one type of features can be digitized at one time. You'll also often have selection functionality in that same `toggleGroup` to avoid digitizing and selecting at the same time.

OTOH, having the spatial operator buttons in that `toggleGroup` makes no sense, since that will effectively remove the selection before having the chance to do the operation.

### DIFF
--- a/src/view/button/SpatialOperatorDifference.js
+++ b/src/view/button/SpatialOperatorDifference.js
@@ -58,8 +58,6 @@ Ext.define('BasiGX.view.button.SpatialOperatorDifference', {
         showSelectMasterSlaveFeatureDialog: true
     },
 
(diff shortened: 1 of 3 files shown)


marcjansen:
Fine for me, but more seasoned developers e.g. @dnlkoch should probably say sth. as well. If this blocks you, we can get it in first, and then probably react to any errors in applications.

hwbllmnn:
I'll merge this then, because it's indeed blocking me ATM. If there are concerns later, I'll be happy to address them.

marcjansen:
Perfectly fine for me.

PR #376, opened 2018-12-10
Add basic instanciation tests and cleanup during tests

marcjansen:
This actually revealed one bug (unchecked setting of `tooltip` viewModel config, which would err out when destroying components). 

Several components would not properly cleanup behind them, so that visual artifacts remained in the karma debug page.

Please review.


marcjansen:
Thanks for the review.

PR #381, opened 2018-12-20
Fix spelling and wording in comments

marcjansen:
Title says it all. Please review.

### DIFF
--- a/src/util/Layer.js
+++ b/src/util/Layer.js
@@ -43,12 +43,12 @@ Ext.define('BasiGX.util.Layer', {
         NAME_MEASURE_LAYER: 'basigx-measure-layer',
 
         /**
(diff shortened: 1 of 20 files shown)


src/util/Layer.js
```diff
@@ -43,12 +43,12 @@ Ext.define('BasiGX.util.Layer', {
         NAME_MEASURE_LAYER: 'basigx-measure-layer',
 
         /**
-         * Get an ol3-layer by the given key-value constellation.
+         * Get an ol-layer by the given key-value constellation.
          *
          * @param {String} key - the layers property name
          * @param {String} val - the layers property value for the given key
```
chrismayer: `layer's` ? (same the line above)

marcjansen: Probably right, I'll fix these in a follow up.

PR #382, opened 2018-12-20
Fix spelling

marcjansen:
Follow-up of #381, adressing remarks by @chrismayer.

Please review.


marcjansen:
Thanks flr the review!

PR #385, opened 2019-01-15
Allow to uncheck text symbolizer in sld styler

LukasLohoff:
Resolves an issue with the BasiGX text symbolizer: The default style couldn't be disabled, which resulted in URLs appearing on the map when the default label attribute `name` was used
- adds a checkbox to the text symbolizer form which disables it
- adds util method to remove the text symbolizer from a given rule
- adds tests for the text symbolizer and the `removeTextSymbolizerFromRule` util method

@terrestris/devs Please review

### DIFF
--- a/src/util/SLD.js
+++ b/src/util/SLD.js
@@ -341,7 +341,7 @@ Ext.define('BasiGX.util.SLD', {
          *
          * @return {Object} sldObject The updated SLD object.
          * @param {String} ruleName The name of the rule that should be updated.
-         * @param {Object} sldObj The sldObj to update, containing the rule.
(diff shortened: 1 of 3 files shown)


src/util/SLD.js
```diff
@@ -336,6 +336,39 @@ Ext.define('BasiGX.util.SLD', {
             return fill;
         },
 
+        /**
+         * Method removes the point symbolizer for a given rule and sldObject
```
hwbllmnn: point -> text

LukasLohoff: Missed that, should be fixed now!

PR #414, opened 2019-03-12
Add a layer preview window

hwbllmnn:
Adds a layer preview window:

* shows the legend, if any
* can zoom to the data extent, if any
* has zoom controls
* has a scale combo
* shows a feature info window when clicking on a feature

### DIFF
--- a/src/view/combo/ScaleCombo.js
+++ b/src/view/combo/ScaleCombo.js
@@ -125,7 +125,7 @@ Ext.define('BasiGX.view.combo.ScaleCombo', {
             me.map = BasiGX.util.Map.getMapComponent().getMap();
         }
         var mapView = me.map.getView();
-        if (this.getUseScalesFromMap) {
+        if (this.getUseScalesFromMap()) {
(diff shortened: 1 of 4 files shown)


src/view/combo/ScaleCombo.js
```diff
@@ -152,6 +157,33 @@ Ext.define('BasiGX.view.combo.ScaleCombo', {
...
+        var maxResolution = view.getMaxResolution();
+        var units = view.getProjection().getUnits();
+        var scales = [];
+        var cur = minResolution;
+        while(cur < maxResolution) {
+            scales.push({
+                scale: resToScale(cur, units).toFixed(0),
+                resolution: cur
+            });
+            cur *= 2;
```
weskamm: Why dont you just use map.getView().getResolutions() and iterate over them instead of guessing here?

hwbllmnn: When I was testing with the SHOGun-webapp `getResolutions()` was null, so I switched to guessing. I can fix this later when I have an application that actually has resolutions (the calculation only takes place when switched on anyway).

PR #424, opened 2019-04-03
Add some basic tests & cleanup on destroy

marcjansen:
This adds a couple of basic tests and also cleans up better after we destroy instances.


marcjansen:
Merci for le review

PR #425, opened 2019-04-04
Make ZoomIn and ZoomOut buttons more configurable

annarieger:
Added following options to make zoom buttons config more comfortable:
* `enableZoomInWithBox` / `enableZoomOutWithBox` -> Enables zoom with box without `shift` key
* `enableAnimation` -> Enables animation while zooming
* `dragZoomInteraction` / `dragZoomOutInteraction` -> Custom `ol.interaction.DragZoom` which will be used while zooming with box
* `animationDuration` -> Custom animation duration

Please review @weskamm

### DIFF
--- a/src/view/button/ZoomIn.js
+++ b/src/view/button/ZoomIn.js
@@ -60,7 +60,12 @@ Ext.define('BasiGX.view.button.ZoomIn', {
     glyph: 'xf00e@FontAwesome',
     html: '<i class="fa fa-search-plus fa-2x"></i>',
 
-    toggleGroup: 'navigation',
+    /**
+     * ZoomIn button is not toggleable per default and behaves like simple
+     * button. If #toggleGroup is set by instantiation the `click` handler will
+     * be ignored and `toggle` handler will be used instead.
+     */
+    toggleGroup: null,
 
     /**
      * A config object to show this tool in action (live demo) when using the
@@ -78,32 +83,43 @@ Ext.define('BasiGX.view.button.ZoomIn', {
     config: {
         /**
          * When set to true zoom in by clicking and dragging on the map is
-         * enabled. Default is true.
+         * enabled. Only applicable if instantiated as toggle button.
+         * Default is true.
          */
         enableZoomInWithBox: true,
         /**
          * Whether zoom action should be animated or not. Default is true.
          */
-        enableAnimation: true,
+        animate: true,
         /**
          * Reference to ol DragZoom interaction which will be used if
          * #enableZoomInWithBox is set to true.
          */
-        dragZoomInteraction: null,
+        dragZoomInInteraction: null,
         /**
          * Default zoom animation duration in milliseconds. Only applicable if
-         * #enableAnimation is set to true.
+         * #animate is set to true.
          */
         animationDuration: 500
     },
 
     listeners: {
-        toggle: function (btn, pressed) {
+        afterrender: function() {
             var me = this;
-            //fallback
             if (Ext.isEmpty(me.olMap)) {
                 me.olMap = BasiGX.util.Map.getMapComponent().getMap();
             }
+        },
+        click: function() {
+            var me = this;
+            // do nothing if configured as toggle button
+            if (!Ext.isEmpty(me.toggleGroup)) {
+                return;
+            }
+            me.zoomIn();
+        },
+        toggle: function (btn, pressed) {
+            var me = this;
             if (me.enableZoomInWithBox) {
                 if (!me.dragZoomInInteraction) {
                     me.dragZoomInInteraction = new ol.interaction.DragZoom({
@@ -137,7 +153,7 @@ Ext.define('BasiGX.view.button.ZoomIn', {
         var olView = me.olMap.getView();
 
         // This if is need for backwards comaptibility to ol
-        if (me.enableAnimation) {
+        if (me.animate) {
             if (ol.animation) {
                 zoom = ol.animation.zoom({
                     resolution: olView.getResolution(),
--- a/src/view/button/ZoomIn.js
+++ b/src/view/button/ZoomIn.js
@@ -35,8 +35,12 @@ Ext.define('BasiGX.view.button.ZoomIn', {
         data: {
             tooltip: 'Hineinzoomen',
             text: null,
-            documentation: '<h2>Hineinzoomen</h2>• Ein Klick auf den Button ' +
-                'vergrößert die Karte um eine Zoomstufe.'
+            documentation: '<h2>Hineinzoomen</h2>' +
+                '• Ein Klick auf den Button aktiviert ZoomIn-Modus:<br>' +
+                '• Ein Klick in die Karte vergrößert sie um eine Zoomstufe. ' +
+                '• Wird ein Rechteck über die Karte gezogen, zoomt die Karte ' +
+                'zum gewählten Ausschnitt (Button muss mit der Option ' +
+                '`enableZoomInWithBox=true` konfiguriert sein).'
         }
     },
 
--- a/src/view/button/ZoomIn.js
+++ b/src/view/button/ZoomIn.js
@@ -56,6 +56,8 @@ Ext.define('BasiGX.view.button.ZoomIn', {
     glyph: 'xf00e@FontAwesome',
     html: '<i class="fa fa-search-plus fa-2x"></i>',
 
+    toggleGroup: 'navigation',
+
     /**
      * A config object to show this tool in action (live demo) when using the
      * context sensitive help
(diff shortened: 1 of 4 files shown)


src/view/button/ZoomIn.js
```diff
@@ -69,37 +75,84 @@ Ext.define('BasiGX.view.button.ZoomIn', {
...
-        handler: function() {
-            var me = this;
-            var olMap = me.olMap;
-            var olView;
-            var zoom;
+        /**
+         * When set to true zoom in by clicking and dragging on the map is
+         * enabled. Default is true.
+         */
+        enableZoomInWithBox: true,
```
marcjansen: Does this change the current default behaviour?

src/view/button/ZoomIn.js
```diff
@@ -69,37 +75,84 @@ Ext.define('BasiGX.view.button.ZoomIn', {
...
+        /**
+         * Whether zoom action should be animated or not. Default is true.
+         */
+        enableAnimation: true,
```
marcjansen: Name suggestion `animate`?

src/view/button/ZoomIn.js
```diff
@@ -69,37 +75,84 @@ Ext.define('BasiGX.view.button.ZoomIn', {
...
+        /**
+         * Reference to ol DragZoom interaction which will be used if
+         * #enableZoomInWithBox is set to true.
+         */
+        dragZoomInteraction: null,
```
marcjansen: Do we expect people to pass this, or is it some slot in which we keep our internal reference?

src/view/button/ZoomOut.js
```diff
@@ -73,33 +80,83 @@ Ext.define('BasiGX.view.button.ZoomOut', {
...
+        enableAnimation: true,
+        /**
+         * Reference to ol DragZoom interaction which will be used if
+         * #enableZoomOutWithBox is set to true.
+         */
+        dragZoomOutInteraction: null,
+        /**
+         * Default zoom animation duration in milliseconds.
+         */
+        animationDuration: 500
```
marcjansen: Can we put these configs in the parent class?

src/view/button/ZoomIn.js
annarieger: Currently  zoom is possible by click into map or by dragging a rectangle while `shift` key is pressed, what is, I think, the default behaviour of ol zoom control.

This config doesn't change it, but provides the additional method to zoom by dragging a rectangle without `shift`

src/view/button/ZoomIn.js
annarieger: If it's not passed, a new `ol.Interaction.DragZoom` will be created and set on button on toggle --> s. [here](https://github.com/terrestris/BasiGX/pull/425/files#diff-cfb68a0905b47011db0dfb25e090a910R108)

weskamm:
I like the changes and agree with the notes of @marcjansen :+1:

weskamm:
As the button now is a togglebutton, we should make sure to keep the old behaviour as default.
Maybe we can keep the old button and just use the togglebutton when the config by user intends it.

annarieger:
> As the button now is a togglebutton, we should make sure to keep the old behaviour as default.
> Maybe we can keep the old button and just use the togglebutton when the config by user intends it.

You're right. I'll take a look at it now.

src/view/button/ZoomIn.js
marcjansen: But default right now is:

* Click -> zoom in
* click again -> zoom in

Is it now:

* Click -> zoom in, toggle button to pressed
* click again -> untoggle pressed?

I see if I can test this.

src/view/button/ZoomIn.js
annarieger: Renamed in 030dc92

src/view/button/ZoomIn.js
annarieger: Original behaviour restored, see also https://github.com/terrestris/BasiGX/pull/425#issuecomment-479861743 from @weskamm and 030dc92

So default behaviour is unchanged now (click on button -> zoom in/out), additional configs like zoom with rectangle without shift are only applicable if button configured as toggle button

src/view/button/ZoomOut.js
annarieger: Zoom buttons have the `BasiGX.view.button.Base` class as parent which is IMO not suitable for these special configs.
I'm not sure if we need one another extra parent class containing shared configs for only two zoom buttons so I leave it unchanged here.

annarieger:
Original behaviour (zoom button = simple button, each click leads to zoom in/out on the map) is restored. Please kindly review once again, @weskamm & @marcjansen

src/view/button/ZoomIn.js
```diff
@@ -69,37 +80,95 @@ Ext.define('BasiGX.view.button.ZoomIn', {
...
+        afterrender: function() {
+            var me = this;
+            if (Ext.isEmpty(me.olMap)) {
+                me.olMap = BasiGX.util.Map.getMapComponent().getMap();
             }
+        },
+        click: function() {
+            var me = this;
+            // do nothing if configured as toggle button
+            if (!Ext.isEmpty(me.toggleGroup)) {
```
chrismayer: I think `toggleGroup` is not the right indicator for this. You can have toggleable button without belonging to a ToggleGroup. [enabeToggle](https://docs.sencha.com/extjs/6.2.0/classic/Ext.button.Button.html#cfg-enableToggle) is the right indicator IMHO.

src/view/button/ZoomOut.js
```diff
@@ -73,33 +85,95 @@ Ext.define('BasiGX.view.button.ZoomOut', {
...
+        afterrender: function () {
+            var me = this;
+            if (Ext.isEmpty(me.olMap)) {
+                me.olMap = BasiGX.util.Map.getMapComponent().getMap();
+            }
+        },
+        click: function () {
+            var me = this;
+            // do nothing if configured as toggle button
+            if (!Ext.isEmpty(me.toggleGroup)) {
```
chrismayer: See comment in `ZoomOut` button.

src/view/button/ZoomOut.js
```diff
@@ -73,33 +85,95 @@ Ext.define('BasiGX.view.button.ZoomOut', {
...
+                    me.dragZoomOutInteraction = new ol.interaction.DragZoom({
+                        condition: ol.events.condition.always,
+                        duration: me.animationDuration,
+                        out: true
+                    });
+                    me.olMap.addInteraction(me.dragZoomOutInteraction);
+                }
+            }
+            if (pressed) {
+                me.olMap.on('click', me.zoomOut, me);
```
chrismayer: Doesn't this conflict with the `ol.interaction.DragZoom` if `pressed` and `enableZoomOutWithBox` are both set to true?

src/view/button/ZoomIn.js
```diff
@@ -69,37 +80,95 @@ Ext.define('BasiGX.view.button.ZoomIn', {
...
+                if (!me.dragZoomInInteraction) {
+                    me.dragZoomInInteraction = new ol.interaction.DragZoom({
+                        condition: ol.events.condition.always,
+                        duration: me.animationDuration
+                    });
+                    me.olMap.addInteraction(me.dragZoomInInteraction);
+                }
+            }
+            if (pressed) {
+                me.olMap.on('click', me.zoomIn, me);
```
chrismayer: Doesn't this conflict with the `ol.interaction.DragZoom` if `pressed` and `enableZoomInWithBox` are both set to true?

src/view/button/ZoomIn.js
annarieger: Nice catch, thanks! Fixed in 5c11570

src/view/button/ZoomOut.js
annarieger: s. above

src/view/button/ZoomOut.js
annarieger: I think it's fine here. Just tested it with a real application - `click` event won't fire if a rectange was drawn per dragging.

Furthermore,  `ol.events.condition.always` which actually always returns `true`, has no effect for single clicks on the map (probably cause bounding box of clicked point is too small to force changing of view resolution -> s. [here](https://github.com/openlayers/openlayers/blob/v4.6.5/src/ol/interaction/dragzoom.js#L74))

src/view/button/ZoomIn.js
annarieger: S. above

src/view/button/ZoomOut.js
chrismayer: Thanks for clarification. This is OK for me then.

src/view/button/ZoomIn.js
chrismayer: Thanks for clarification. This is OK for me then.

annarieger:
Thanks for reviews and remarks, guys. Merging now :pineapple:

PR #436, opened 2019-04-26
Don't try to instantiate a non-existing global

marcjansen:
This PR suggests to only call `new Jsonix.Context(…)` when all globals could be found on the page.

Please review.


marcjansen:
Thanks for the review.

PR #437, opened 2019-04-29
Add missed scope on map click listener

annarieger:
This fixes the `zoomOut` button behavior to ensure that `click` listener on the map turns unregistered if button was untoggled.


annarieger:
Thanks :pineapple:

PR #438, opened 2019-05-02
Adjustments to the WFS utils

mholthausen:
Adjustments to WFS utils, in particular for determining the `namespaceUri` dynamically in `executeWfsGetFeature`

### DIFF
--- a/src/util/WFS.js
+++ b/src/util/WFS.js
@@ -24,7 +24,8 @@ Ext.define('BasiGX.util.WFS', {
         'BasiGX.util.CSRF',
         'BasiGX.util.Url',
         'BasiGX.util.Filter',
-        'BasiGX.util.Jsonix'
+        'BasiGX.util.Jsonix',
+        'BasiGX.util.Namespace'
     ],
 
     inheritableStatics: {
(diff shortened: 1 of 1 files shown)


src/util/WFS.js
```diff
@@ -487,6 +488,8 @@ Ext.define('BasiGX.util.WFS', {
 
             var featureType = layer.getSource().getParams().LAYERS;
```
marcjansen: Unrelated to the changes in this specific PR, what will happen if the `featureType` varuiable is e.g. `foo:bar,humpty:dumpty`?

chrismayer: Also relying on the uppercase notation for `LAYERS` is not 100% fail safe. A lot of WMS / WFS servers also accept the lowercase variant `layers` and OL just offers in `getParams` what has been configured. To be honest not sure what the standard says about upper-/lowercase.
I usually try `var ft = layer.getSource().getParams().LAYERS || layer.getSource().getParams().layers`

marcjansen: See #439, @chrismayer

PR #439, opened 2019-05-04
Get values for keys with unclear spelling in objects

marcjansen:
It sometimes is desirable to get values from objects where the actual spelling of the key is unclear, e.g. when we write stuff like…

```js
var params = wmsSource.getParams();
var layer = params.LAYERS // will miss params.layers and any other spelling
```

Since the above reflects user-passed values and they are accepted and vald in basically any casing, but at the same time several spelling are more common than others, this PR adds a new method to do just this.

```js
var params = wmsSource.getParams();
var layers = BasiGX.util.Object.getValueSpellingVariants(
    params, ['LAYERS', 'layers'], /^layers$/i
);
// layers now contains the value of e.g. params.LAYERS,
// params.layers or params.LaYeRs, whatever we find first
```

Since getting the layers param is quite often requested, this also adds anothe rconvenience method `layersFromParams`, that calls into `getValueSpellingVariants` with the above configuration.

Please review.

See also https://github.com/terrestris/BasiGX/pull/438#discussion_r280983349

### DIFF
--- a/src/util/Object.js
+++ b/src/util/Object.js
@@ -97,6 +97,24 @@ Ext.define('BasiGX.util.Object', {
         },
 
         /**
+         * A utility method to get the layers-key from an object which
+         * represents URL params. Will first try the keys `LAYERS`, then
+         * `layers` and finally all sorts of mixed casing spellings like
+         * `lAyERs`.
+         *
+         * @param {Object} params The object to look in.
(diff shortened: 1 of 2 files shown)


marcjansen:
Thanks for the reviews, @chrismayer & @dnlkoch. I'll merge when green.

PR #440, opened 2019-05-04
Use configured animation duration and fix several other minors in zoom buttons

marcjansen:
This fixes a bunch of smaller issues and one bug in the `ZoomIn` and `ZoomOut` buttons:

* The `documentation` entries got some additional markup so that at least they look like a list even though they aren't one
* In the API docs for some configs some sentences were reworded, so that new lines in `/** */` would not start with `#`; that leads to rendering issues once the documentation is rendered with JSDuck (the picture below shows the current rendering before the changes in this PR)
a
* One typo in inline docs is removed
* bugfix: A hardcoded animation duration of 500 ms is removed and instead the already existing user configurable value is used.

Please review.

### DIFF
--- a/src/view/button/ZoomIn.js
+++ b/src/view/button/ZoomIn.js
@@ -154,7 +154,7 @@ Ext.define('BasiGX.view.button.ZoomIn', {
         var zoom;
         var olView = me.olMap.getView();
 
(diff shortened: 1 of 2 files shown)


marcjansen:
Thanks for the review.

PR #441, opened 2019-05-05
Use layersFromParams utility method

marcjansen:
This PR suggests to use `BasiGX.util.Object.layersFromParams` from #439 in two classes:

* `BasiGX.util.WFS` (where this was [first noticed](https://github.com/terrestris/BasiGX/pull/438#discussion_r280983349))
* `BasiGX.util.Layer`

The method `BasiGX.util.Layer.getLayerByLayersParam`was also changed to find not only layers with a source `ol.source.TileWMS`, but als those with a source of `ol.source.ImageWMS`. As that partical method also didn't have any tests, I added some as well.

Please review.

### DIFF
--- a/src/util/Layer.js
+++ b/src/util/Layer.js
@@ -21,7 +21,8 @@
  */
 Ext.define('BasiGX.util.Layer', {
     requires: [
-        'BasiGX.util.Map'
+        'BasiGX.util.Map',
+        'BasiGX.util.Object'
     ],
     statics: {
         /**
@@ -92,6 +93,7 @@ Ext.define('BasiGX.util.Layer', {
             if (!map) {
                 map = BasiGX.util.Map.getMapComponent().getMap();
             }
+            var ObjectUtil = BasiGX.util.Object;
             var mapLayers = map.getLayers();
             var olLayer;
             var foundIt = false;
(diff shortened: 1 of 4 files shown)


test/spec/util/Object.test.js
```diff
@@ -6,35 +6,127 @@ describe('BasiGX.util.Object', function() {
...
-            var retVal1 = BasiGX.util.Object.getValue('level', testObject);
-            expect(retVal1).to.be('first');
+            it('returns `undefined` when neither commonVariants nor regex passed', function() {
+                expect(cls.getValueSpellingVariants({foo: 1})).to.be(undefined);
+            });
 
-            var retVal2 = BasiGX.util.Object.getValue('firstNested/level',
-                testObject);
-            expect(retVal2).to.be('second');
+            it('returns `undefined` when neither commonVariants nor regex passed', function() {
```
chrismayer: duplicate to L17ff?

marcjansen: Yes, nice catch.

marcjansen:
This includes a circular dependency… investigating later

marcjansen:
* `BasiGX.util.Layer` depends on
  * `BasiGX.util.Object` (since this PR) which depends on
    * `BasiGX.util.Application` (since long ago), which depends on
      * `BasiGX.util.Layer` :disappointed:

marcjansen:
@chrismayer (or someone else): I am very interested in your thoughts on 6fec466. It would be very kind if you could review again. TIA.

src/util/Layer.js
```diff
@@ -101,10 +103,14 @@ Ext.define('BasiGX.util.Layer', {
...
                 if (layer instanceof ol.layer.Layer &&
-                    layer.getSource() instanceof ol.source.TileWMS &&
-                    layer.getSource().getParams().LAYERS === layersParam) {
-                    olLayer = layer;
-                    foundIt = true;
+                    (layer.getSource() instanceof ol.source.TileWMS
+                    || layer.getSource() instanceof ol.source.ImageWMS)) {
+
+                    var params = layer.getSource().getParams();
+                    if(ObjectUtil.layersFromParams(params) === layersParam) {
```
dnlkoch:
```suggestion
                    if (ObjectUtil.layersFromParams(params) === layersParam) {
```

marcjansen:
Thanks for the reviews

PR #446, opened 2019-05-14
Add proper geometry comparison

hwbllmnn:
Adds a geometry comparison that compares with a delta. Also fixes a lot of linting errors (unsure why travis doesn't find them, they fail my tests locally).

### DIFF
--- a/src/util/Geometry.js
+++ b/src/util/Geometry.js
@@ -21,6 +21,208 @@ Ext.define('BasiGX.util.Geometry', {
     statics: {
 
         /**
+         * Check the features geometries' for equality and return the features
+         * which have duplicates.
+         *
+         * @param {ol.Feature[]} features the features to check
+         * @param {number} delta the delta for coordinate comparison, default
+         * is 0.000001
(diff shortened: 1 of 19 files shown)


weskamm:
Very nice, thanks!

marcjansen:
The linting errors usually show up, if there is another config (even higher up in the hierarchy) that tales effect. I suppose this was started from a lib folder / git submodule in a project. Becaise AFAICT we do not enforce e.g. one quote style.

src/util/Geometry.js
```diff
@@ -20,6 +20,208 @@ Ext.define('BasiGX.util.Geometry', {
...
+
+        /**
+         * Checks two point geometries for equality.
+         *
+         * @param {ol.geom.Point} p1 the first point
+         * @param {ol.geom.Point} p2 the second point
+         * @param {number} delta the delta for comparison
+         * @return {boolean} true if equal
+         */
+        isDuplicatePoint: function(p1, p2, delta) {
```
marcjansen: These all only take into account two dimensions, which is perfectly fine, but we should document this IMO.

PR #447, opened 2019-05-15
Fix namespace util backwards compatibility

hwbllmnn:
Fixes the namespace/wfs utils backwards compatibility. Also addresses comments from #446 .


weskamm:
:+1:

PR #448, opened 2019-05-16
Add minified version on github pages

weskamm:
title says it all


package.json
```diff
@@ -67,6 +67,7 @@
         "puppeteer": "1.16.0",
         "rimraf": "2.6.3",
         "serve": "11.0.0",
-        "sinon": "7.3.2"
+        "sinon": "7.3.2",
+        "uglify-js": "3.5.12"
```
marcjansen: No need to add it to `package.json` if you use `npx`

PR #456, opened 2019-05-27
More flexible regular expression to work w/ filters

marcjansen:
This changes the WFS util so it can be used to work with filters
that either have or don't have the `ogc:` namespace prefix.

The methods `unwrapFilter` and `getFilterPrefix` both will be
positively affected by this change.

Please review.


marcjansen:
Cheers for the review.

PR #474, opened 2019-07-11
Add a CSW form to retrieve WMS layers through CSW

weskamm:
This is currently restricted to WMS services only, but should be easily extendible to support WFS.

I tried hard to parse the CSW responses with Jsonix, but i failed and decided to use the DOMParser for this trivial task.

All it does is to request the given csw endpoint with `GetRecords`. It collects all records that have a dc:URI with an OGC:WMS protocol and generates buttons for that.
When clicking a button, the well known `BasiGX.view.form.AddWms` is used to retrieve the layer by automatically issuing a GetCapabilities request against the URL.

image

### DIFF
--- a/src/view/form/CSW.js
+++ b/src/view/form/CSW.js
@@ -1,0 +1,614 @@
+/* Copyright (c) 2015-present terrestris GmbH & Co. KG
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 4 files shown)


src/view/form/CSW.js
```diff
@@ -0,0 +1,614 @@
...
+         * The name of this parameter is taken from the config option of
+         * `Ext.Ajax`, turning the boolean logic around would be even more
+         * confusing.
+         *
+         * Defaults to `true`, e.g. the cache buster will be send along in the
+         * GET-request.
+         *
+         * @type {Boolean}
+         */
+        disableCaching: true
```
dnlkoch: Can we have something like `additionalRequestParams`(with the current opts `useDefaultXhrHeader` and `disableCaching` as defaults)  instead? So we would support all available `Ext.Ajax.request()` config options.

PR #489, opened 2019-08-13
Execute CSW search on ENTER in search text field

chrismayer:
This ensures that the CSW search is executed if the ENTER key is hit inside the search text field of the `BasiGX.view.form.CSW`.


chrismayer:
Tests are failing because of `No binary for ChromiumHeadless browser on your platform. Please, set "CHROMIUM_BIN" env variable.`.

chrismayer:
Thanks for the review @weskamm, I'll merge now.

PR #492, opened 2019-08-14
Fix test failures

weskamm:
Trying to fix karma issues with browser env variable


chrismayer:
Chromium seem to be no more installed in the Travis environment, seehttps://travis-ci.community/t/no-binary-for-chromiumheadless-browser-on-your-platform/1858

weskamm:
Thanks for the info, we will switch to chrome then...

PR #497, opened 2019-08-22
Wfs util: use passed feature type if provided

ahennr:
If `featureType` is passed as parameter, its determination from `PARAMS` is omitted.


ahennr:
THX for review @hwbllmnn

PR #504, opened 2019-09-05
Support Vector Tile layers in MapLoadingStatusBar

chrismayer:
This ensures Vector Tile layers are also taken into account for the loading status in the `MapLoadingStatusBar`. Therefore all sources inheriting from `ol.source.Tile` (not only `ol.source.TileImage`) will be checked.


chrismayer:
Thanks for the review @marcjansen

marcjansen:
We should be aware that the type hierarchy for newer versions of OpenLayers has changed and we will likely need to adapt to this sooner or later.

PR #507, opened 2019-09-09
Fixes tooltip binding in base button

KaiVolland:
This removes the constructor and adds binding for `text` and `tooltip` for the `BasiGX.view.button.Base`.

### DIFF
--- a/src/view/button/Base.js
+++ b/src/view/button/Base.js
@@ -21,21 +21,23 @@
 Ext.define('BasiGX.view.button.Base', {
     extend: 'Ext.Button',
     xtype: 'basigx-button-base',
+
     viewModel: {
         data: {}
     },
     bind: {},
+
     /**
      *
      */
-    constructor: function() {
+    constructor: function () {
         var me = this;
         me.callParent(arguments);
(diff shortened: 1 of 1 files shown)


KaiVolland:
Ok. Thakns for the review and hint. No i haven't checked this with the modern toolkit. But it didn't work for me with the classic. I may just add the bindings and leave the constructor.

weskamm:
If i remember correctly there was a setter missing in one of the toolkits leading to errors. Maybe it has been added in the meantime, i dont know

KaiVolland:
Actually there is a `setTooltip` method for the Button in modern too now: https://docs.sencha.com/extjs/6.2.0/modern/Ext.Button.html#method-setTooltip

As the basebutton does nothing but the hack for the tooltip the whole component could be removed IMHO.

KaiVolland:
For now i just change the binding to actually "bind" to the tooltip property in the viewmodel and not just set the tooltip-value. I'm surprised that this worked before.

weskamm:
This looks strange ( and it did before), but if it works feel free to merge...

PR #513, opened 2019-09-24
Reset window reference on destroying properly

annarieger:
Previously `AddWms` and `CoordinateTransform` windows could be shown only once. 

After the component was called repeatedly (e.g. after the window was closed), the console error was thrown, cause the window instance was not reset properly:
image

Please review @weskamm


weskamm:
looks good but please fix tests

annarieger:
Tests are fixed, thanks for the review :pineapple:

PR #519, opened 2019-10-01
Replace maxFeatures string by an empty one if maxFeatures is not defined

ahennr:
This PR affects the `WPSUtil`: If GetFeature is used as referenced datatype and `maxFeature` is not set its occurrence in request string must be removed.


buehner:
:+1:

PR #540, opened 2019-11-18
Make hoverColor property of hover plugin configurable

annarieger:
Adds possibility to set custom color while hovering over features.

The default behaviour (red color with 0.6 opacity) is unchanged.

Please review @weskamm


weskamm:
:+1:

PR #548, opened 2019-11-28
Make use of custom layer search configuration with WFS search

annarieger:
## Support for querying of numerical attributes
Previously only `PropertyIsLike`  filter was implemented so only string based attributes could be searched through via WFS. 
This MR extends `MutiSearchWFSSearchGrid` class and adds possibility to query numeric attributes (`xsd:number` or `xsd:int`) as well.

To achieve this, an additional filter function `stringFormat` was implemented.

Please note, that numeric attributes can be only be queried and filtered, if an additional extension [`terrestris-filterfunctions-0.0.1.jar`](https://nexus.terrestris.de/repository/public/de/terrestris/terrestris-filterfunctions/0.0.1/terrestris-filterfunctions-0.0.1.jar) is available in geoserver. Please refer to https://github.com/terrestris/terrestris-filterfunctions for further details. 

## Support for certain layer attributes
If layer is configured with custom property `searchable` and has a non empty array of `searchColumns` as further config defined, only these columns will be queried via `GetFeature` request. Otherwise all layer attributes will be used while searching (default behaviour)

## Support for certain layer search results template
If layer is configured with a custom property `searchable` and has a further non empty property `searchTemplate`, this template will be used to represent found features inside of search results grid.
Otherwise the attribute, which contains the search term will be shown as result title instead (default behaviour).

Also introduced `StringTemplate` util and moved `getHoverTextFromTemplate` method of `Hover.js` class inside, since this method can be used for both hover and WFS search results.

### DIFF
--- a/src/plugin/Hover.js
+++ b/src/plugin/Hover.js
@@ -19,6 +19,10 @@
 Ext.define('BasiGX.plugin.Hover', {
     extend: 'Ext.plugin.Abstract',
 
+    requires: [
+        'BasiGX.util.StringTemplate'
+    ],
+
     alias: 'plugin.hover',
     pluginId: 'hover',
 
@@ -587,26 +591,29 @@ Ext.define('BasiGX.plugin.Hover', {
         var me = this;
         var innerHtml = '';
         var hoverfieldProp = me.self.LAYER_HOVERFIELD_PROPERTY_NAME;
+        var templateUtil = BasiGX.util.StringTemplate;
+        var templateConfig = {
+            prefix: me.self.HOVER_TEMPLATE_PLACEHOLDER_PREFIX,
+            suffix: me.self.HOVER_TEMPLATE_PLACEHOLDER_SUFFIX
+        };
 
-        Ext.each(features, function(feat) {
-            var layer = feat.get('layer');
-            var hoverFieldProp = layer.get(hoverfieldProp);
-
-            var hoverText = me.getHoverTextFromTemplate(feat, hoverFieldProp);
-
+        Ext.each(layers, function (layer) {
             innerHtml += '<b>' + layer.get('name') + '</b>';
+            Ext.each(features, function(feat) {
+                if (feat && feat.get('layer') === layer) {
+                    var hoverFieldProp = layer.get(hoverfieldProp);
 
-            // we check for existing feature here as there maybe strange
-            // situations (e.g. when zooming in unfavorable situations) where
-            // feat is undefined
-            if (feat) {
-                if (layer.get('type') === 'WFSCluster') {
-                    var count = feat.get('count');
-                    innerHtml += '<br />' + count + '<br />';
-                } else {
-                    innerHtml += '<br />' + hoverText + '<br />';
+                    var hoverText = templateUtil.getTextFromTemplate(
+                        feat, hoverFieldProp, templateConfig);
+
+                    if (layer.get('type') === 'WFSCluster') {
+                        var count = feat.get('count');
+                        innerHtml += '<br />' + count + '<br />';
+                    } else {
+                        innerHtml += '<br />' + hoverText + '<br />';
+                    }
                 }
-            }
+            });
         });
 
         return innerHtml;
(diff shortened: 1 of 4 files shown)


src/view/grid/MultiSearchWFSSearchGrid.js
```diff
@@ -1,3 +1,4 @@
+/* eslint max-len: ["error", { "comments": 100 }] */
```
hwbllmnn: We should definitely increase this globally, perhaps even to 120...

src/view/grid/MultiSearchWFSSearchGrid.js
```diff
@@ -439,31 +449,76 @@ Ext.define('BasiGX.view.grid.MultiSearchWFSSearchGrid', {
...
+                        // functions list.
+                        // To get this filter work, the additional geoserver
+                        // extension `terrestris-filterfunctions` must be
+                        // installed (see
+                        // https://github.com/terrestris/terrestris-filterfunctions
+                        // for further details)
+                        comparisonFilter =
+                            '<ogc:PropertyIsLike wildCard="*" singleChar="."' +
+                                ' escape="\\" matchCase="false">' +
+                                '<ogc:Function name="stringFormat">' +
```
hwbllmnn: :ok_hand:

src/view/grid/MultiSearchWFSSearchGrid.js
```diff
@@ -498,17 +598,38 @@ Ext.define('BasiGX.view.grid.MultiSearchWFSSearchGrid', {
...
+                            return l.get('name') === ftName;
+                        });
+                    }
+                    if (layer) {
+                        useCustomTemplate = layer && layer.get('searchable') &&
+                            layer.get('searchTemplate');
+                    }
+
                     var displayfield;
+                    var featuretype = feature.id.split('.')[0];
```
hwbllmnn: Note that this will also only work in Geoserver (or other WFS which use IDs like this)

annarieger: There are some more occurrences of this kind of check across the library.. maybe we should refactor it later

src/view/grid/MultiSearchWFSSearchGrid.js
annarieger: :crown:

src/view/grid/MultiSearchWFSSearchGrid.js
marcjansen: Heresy!

src/plugin/Hover.js
```diff
@@ -791,87 +798,6 @@ Ext.define('BasiGX.plugin.Hover', {
...
-
-    /**
-     * Returns the hover text for the passed `feature` and `hoverTemplate`.
-     *
-     * @param {ol.Feature} feature The feature that was hovered.
-     * @param {String} hoverTemplate The template to fill with feature
-     *     attributes.
-     * @return {String} The text for the hovered feature.
-     */
-    getHoverTextFromTemplate: function(feature, hoverTemplate) {
```
marcjansen: Couldn' t we call the method in `BasiGX.util.StringTemplate` and be backwards compatible? I would very much prefer to have that. TIA

src/view/grid/MultiSearchWFSSearchGrid.js
```diff
@@ -439,31 +449,76 @@ Ext.define('BasiGX.view.grid.MultiSearchWFSSearchGrid', {
...
+                                    '*' + me.searchTerm + '*' +
+                                '</ogc:Literal>' +
+                            '</ogc:PropertyIsLike>';
+                        break;
+                    // TODO add support for xsd:date
+                    case 'xsd:int':
+                    case 'xsd:number':
+                        var type = 'java.lang.Double';
+                        // Creates custom filter function `stringFormat` which
+                        // doesn't oficially contained in geoserver filter
```
marcjansen: ~~doesn't oficially~~ …isn't officially…

src/view/grid/MultiSearchWFSSearchGrid.js
marcjansen: Hmmm. I am unsure whether I like this. Does this mean people can only use this grid (when filtering a xsd:number fields) if they have a GeoServer extension installed? If so, we need to have this configurable.

test/spec/util/StringTemplate.test.js
```diff
@@ -0,0 +1,81 @@
+Ext.Loader.syncRequire(['BasiGX.util.StringTemplate']);
+
+describe('BasiGX.util.StringTemplate', function () {
```
marcjansen: Much :heart: for adding these tests :pineapple: !!!

src/plugin/Hover.js
annarieger: I' m not sure, if I understand you.. Do you mean, we should reintroduce the method in `Hover.js` again and do something like this inside:
```
getHoverTextFromTemplate: function(feature, hoverTemplate) {
     BasiGX.util.String.Template(feature, hoverTemplate);
  }
```

src/view/grid/MultiSearchWFSSearchGrid.js
annarieger: will be fixed in follow up soon

src/view/grid/MultiSearchWFSSearchGrid.js
annarieger: Yes, this assumes that a geoserver layer will be requested and geoserver itself has the extension installed.
The approach is still pretty raw and can be surely enhanced later.

For now I don't know, which kind of config could be used here - should we check, if the extension is installed? if geoserver is used? something else?

weskamm: cant we use somethin like `parseFloat` here as a default and if the user configured this classs with `useGeoServerStringExtension` with true - the code below?

hwbllmnn: Not really. The underlying problem is that GeoServer can't handle `LIKE` queries on numeric fields. Manually converting numeric fields to `String` using e.g. `strTrim` or others causes `double` values to possibly be converted to scientific notation, thus rendering the `LIKE` query useless, that's why we invented the `stringFormat` function.

annarieger: > cant we use somethin like `parseFloat` here as a default and if the user configured this classs with `useGeoServerStringExtension` with true - the code below?

This additional config can be surely introduced. There is a bundle of string functions in geoserver, but unforunately there is any function which converts provided argument to the formatted string with specific format. As fallback the only one suitable geoserver filter function `strTrim` can be used here. 

However, the disadvantage of this function is that the long numbers will be converted to it's stringified scientific notation, so queries with `LIKE` will return potentially less results as expected.

Example:
string representation of  `2375239000` is `2.375239e+9`
string representation of `2391101900` is `2.3911019e+9`

While querying with `attr ILIKE "%239%"` only the first match attribut will be returned and the second match is lost.

PR #551, opened 2019-12-03
Handle hovering over vector layers correctly

buehner:
In addition to https://github.com/terrestris/BasiGX/pull/527 this change fixes some strange behaviour when hovering over vector layers.

First time hovering over a vector was fine, but every subsequent hover action on the same feature resulted in a growing layer (and also feature) list in the popup. The problem also occured at locations where two or more features had the same (or at least similar) geometry.

To avoid the growing layer list, there is a simple check whether the layer is already contained in the `hoverLayers` array.
To avoid the growing feature (template/attribute) list of a layer, the (maybe generated) `id` of `feat` is also applied to its clone to be able to check whether this feature should be added in the next iteration or not.

### DIFF
--- a/src/plugin/Hover.js
+++ b/src/plugin/Hover.js
@@ -389,10 +389,19 @@ Ext.define('BasiGX.plugin.Hover', {
                             // from the clusterStyle
                             hvl.setStyle(me.highlightStyleFunction);
                         }
-                        hoverLayers.push(layer);
+                        if(!Ext.Array.contains(hoverLayers, layer)) {
+                            hoverLayers.push(layer);
+                        }
                         if (feat.get('layer') === layer) {
                             var clone = feat.clone();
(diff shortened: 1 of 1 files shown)


src/plugin/Hover.js
```diff
@@ -389,10 +389,19 @@ Ext.define('BasiGX.plugin.Hover', {
                             // from the clusterStyle
                             hvl.setStyle(me.highlightStyleFunction);
                         }
-                        hoverLayers.push(layer);
+                        if(!Ext.Array.contains(hoverLayers, layer)) {
```
hwbllmnn:
```suggestion
                        if (!Ext.Array.contains(hoverLayers, layer)) {
```

buehner: done

PR #555, opened 2019-12-13
Add check for geoserver filter extension used with WFS search

annarieger:
## FEATURE | BUGFIX

Followup of https://github.com/terrestris/BasiGX/pull/548

Introduced an additional config `useGeoServerStringExtension` (default to `false`) to decide, whether the custom geoserver filter extension should be used for attributes filter while querying of numeric attributes.

if not, `strTrim` geoserver filter [string function](https://docs.geoserver.org/stable/en/user/filter/function_reference.html#string-functions) will be used as fallback.

### DIFF
--- a/src/view/form/field/MultiSearchCombo.js
+++ b/src/view/form/field/MultiSearchCombo.js
@@ -93,7 +93,7 @@ Ext.define('BasiGX.view.form.field.MultiSearchCombo', {
 
         /**
          * Whether to use custom geoserver filter function `stringFormat` which
-         * isn't oficially contained in geoserver filter functions list.
+         * isn't officially contained in geoserver filter functions list.
          *
          * Geoserver cannot handle LIKE queries on numeric fields out of the
          * box. For manually converting to string e.g. `strTrim` filter function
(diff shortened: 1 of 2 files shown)


src/view/form/field/MultiSearchCombo.js
```diff
@@ -91,6 +91,27 @@ Ext.define('BasiGX.view.form.field.MultiSearchCombo', {
 
         searchLayerBlackList: [],
 
+        /**
+         * Whether to use custom geoserver filter function `stringFormat` which
+         * isn't oficially contained in geoserver filter functions list.
```
hwbllmnn:
```suggestion
         * isn't officially contained in geoserver filter functions list.
```

src/view/form/field/MultiSearchCombo.js
```diff
@@ -91,6 +91,27 @@ Ext.define('BasiGX.view.form.field.MultiSearchCombo', {
...
+         *
+         * Geoserver cannot handle LIKE queries on numeric fields out of the
+         * box. For manually converting to string e.g. `strTrim` filter function
+         * can be used though. In some cases double values will be possibly
+         * converted to scientific notation, what makes LIKE query useless.
```
hwbllmnn:
```suggestion
         * converted to scientific notation, what makes LIKE queries useless.
```

PR #556, opened 2019-12-13
Add close trigger to `MultiSearchCombo` component

annarieger:
Previously the search result container could be only removed if the search field was cleared.

Added a trigger which allows to hide and show results again.

image

Also added 2 further configs to `MultiSearchWFSSearchGrid` component:
* `zoomToSearchResults` - whether the map should zoom to clicked search result, default is false (old behaviour)
* `zoomToScale` - optional config which defines scale value, which should be used as max zoom by click on search result if `zoomToSearchResults` option is set to true.

### DIFF
--- a/src/view/form/field/MultiSearchCombo.js
+++ b/src/view/form/field/MultiSearchCombo.js
@@ -169,6 +169,27 @@ Ext.define('BasiGX.view.form.field.MultiSearchCombo', {
             handler: function() {
                 this.showSettingsWindow();
             }
+        },
+        close: {
+            cls: 'multisearch-open-container-trigger',
+            handler: function (combo, trigger) {
+                this.changeContainerVisibility(combo, trigger);
+            }
+        }
+    },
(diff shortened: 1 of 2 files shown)


src/view/form/field/MultiSearchCombo.js
```diff
@@ -169,6 +169,27 @@ Ext.define('BasiGX.view.form.field.MultiSearchCombo', {
             handler: function() {
                 this.showSettingsWindow();
             }
+        },
+        close: {
+            cls: 'multisearch-open-container-trigger',
+            handler: function (combo, trigger) {
+                this.changeContainerVisibility(combo, trigger);
```
dnlkoch: Second argument `trigger` isn't defined in the signature of the method.

annarieger: Nice catch :eyes: Removed it in 374fd4c

PR #558, opened 2019-12-16
Add proxy URL config to add wms form

weskamm:
This adds a config option to specify a proxy URL which will be applied to the capabilites requests.

Useful e.g. when having issues with CORS


weskamm:
Thanks, merging now!

PR #560, opened 2019-12-18
Fix CSW window bug

buehner:
Same problem that we had here: https://github.com/terrestris/BasiGX/pull/513

This PR solves the problem


buehner:
thx guys

PR #567, opened 2020-01-15
Fix Layer Preview Window: legend button, tooltips

hblitza:
- move the show legend button from the header to the map
- change icon to _fa-info_
- force legend labels (without labels it makes no sense)
- fix of button tooltips

### DIFF
--- a/src/view/window/PreviewWindow.js
+++ b/src/view/window/PreviewWindow.js
@@ -21,6 +21,14 @@
 Ext.define('BasiGX.view.window.PreviewWindow', {
     extend: 'Ext.window.Window',
 
+    viewModel: {
+        data: {
+            showLegendTooltip: 'Show Legend',
+            zoomToProjectionTooltip: 'Zoom to extent of projection',
+            zoomToLayerTooltip: 'Fit to extent'
+        }
+    },
+
     closeAction: 'hide',
 
     constrain: true,
@@ -32,13 +40,6 @@ Ext.define('BasiGX.view.window.PreviewWindow', {
         dataExtent: [[10000, 20000], [20000, 40000]]
     },
 
-    tools: [{
-        type: 'gear',
-        handler: function() {
-            this.up('window').showLegend();
-        }
-    }],
-
     /**
      * Sets up the map and the controls.
      */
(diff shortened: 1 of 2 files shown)


src/view/window/PreviewWindow.js
```diff
@@ -32,24 +32,38 @@ Ext.define('BasiGX.view.window.PreviewWindow', {
...
      */
     initComponent: function() {
         this.callParent();
         var markup = '<i class="fa fa-globe" aria-hidden="true"></i>';
         var worldIcon = Ext.dom.Helper.createDom(markup);
+
+        /**
+        * Create custom button to show legend.
+        */
+        var buttonInfo = document.createElement('button');
```
marcjansen: Should we use `Ext.dom.Helper.createDom` and friends?

src/view/window/PreviewWindow.js
```diff
@@ -178,11 +192,12 @@ Ext.define('BasiGX.view.window.PreviewWindow', {
...
             version: '1.1.1',
             format: 'image/png',
-            layer: this.layerConfig.source.layerNames
+            layer: this.layerConfig.source.layerNames,
+            legend_options: 'forceLabels:on'
         };
         url += '?' + Ext.Object.toQueryString(params);
         Ext.create('Ext.window.Window', {
-            title: this.getTitle(),
+            title: 'Legend',
```
marcjansen: Why is this needed?

src/view/window/PreviewWindow.js
```diff
@@ -32,24 +32,38 @@ Ext.define('BasiGX.view.window.PreviewWindow', {
...
+        */
+        var buttonInfo = document.createElement('button');
+        buttonInfo.title = 'Show legend';
+        buttonInfo.type= 'button';
+        buttonInfo.innerHTML = '<i class="fa fa-info" aria-hidden="true"></i>';
+
+        var showLegendWindow = function() {
+            this.showLegend();
+        };
+        buttonInfo.addEventListener('click', showLegendWindow.bind(this));
```
weskamm: Cant we just call and bind `this.showLegend` here?

src/view/window/PreviewWindow.js
hblitza: As the title is already set and present in the preview window, this is double-dipping. The simple title 'legend' makes it easier to me. There can be just one legend window at the time, so ne need for the layer-title here. But other opinions are welcome.

marcjansen: This way it will be impossible to translate this title. When this is coming from a configuration it is translatable AFAICT

src/view/window/PreviewWindow.js
marcjansen: Yes, I think so (untested)

 buttonInfo.addEventListener('click', showLegendWindow.bind(this));

```suggestion
        buttonInfo.addEventListener('click', this.showLegend.bind(this));
```

src/view/window/PreviewWindow.js
hblitza: right, I will leave it to this.getTitle and will explore i18n possibilities in Basigx.

weskamm:
Could you please move all translatable texts to the viewmodel - data, like seen e.g. here:

https://github.com/terrestris/BasiGX/blob/bb9f1bba85b81d74cd14bb6d12eac196658929ea/src/view/window/MergeWindow.js#L28 ?

This will make it easier to translate the texts if needed.
Else this looks good to me!

PR #571, opened 2020-01-24
Fix of hover-effect for graphicpool

hblitza:
- makes window size configurable for the graphicpool
- adds _overItemCls_ config to define a hover-css class for each graphic-item in the pool

### DIFF
--- a/src/view/panel/GraphicPool.js
+++ b/src/view/panel/GraphicPool.js
@@ -59,16 +59,6 @@ Ext.define('BasiGX.view.panel.GraphicPool', {
     },
 
     /**
-     * a default width for this component
-     */
-    width: 535,
-
-    /**
-     * a default height for this component
-     */
-    height: 600,
-
-    /**
      * the layout to use
      */
     layout: 'vbox',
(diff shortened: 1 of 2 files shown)


src/view/panel/GraphicPool.js
```diff
@@ -127,7 +130,10 @@ Ext.define('BasiGX.view.panel.GraphicPool', {
         // UI displaying the user pictures
         me.pictureView = Ext.create('BasiGX.view.view.GraphicPool', {
             title: false,
-            backendUrls: me.getBackendUrls()
+            backendUrls: me.getBackendUrls(),
+            overItemCls: me.getOverItemCls(),
+            height: me.height,
```
LukasLohoff:
```suggestion
            height: me.getHeight(),
```
Ext auto-generates getters for config values, I think it's better to use them instead of directly accessing the value

src/view/panel/GraphicPool.js
```diff
@@ -127,7 +130,10 @@ Ext.define('BasiGX.view.panel.GraphicPool', {
...
+            width: me.width
```
LukasLohoff:
```suggestion
            width: me.getWidth()
```

Issue #503, opened 2019-09-02
Wrong version number published to npm


dnlkoch:
Using the current [release script](https://github.com/terrestris/BasiGX/blob/master/package.json#L49) will lead to a release with an invalid version number on npmjs.com.

Reason: Because the target files will be copied to the `dist` folder **before** the version number will be increased by `np`, the package will be released with the preexisting verison number.

Note: The release of version 1.1.0 has been fixed manually.

LukasLohoff:
Right now the npm release script doesn't work at all. It fails with 

```
  ✔ Prerequisite check
  ✔ Git
  ↓ Cleanup [skipped]
  ✔ Prerequisite check
  ✔ Git
  ↓ Cleanup [skipped]
  ✔ Installing dependencies using npm
  ✔ Running tests using npm
  ✔ Bumping version using npm
  ✖ Publishing package using npm
    → The project was rolled back to its previous state.
    Pushing tags
    Creating release draft on GitHub

✖ Error publishing package:
Command failed with exit code 1: npm publish dist
```

To do a manual release we currently have to:

- bump the version in package.json and commit & push the changes
- draft a new release on https://github.com/terrestris/BasiGX/releases
- change the version in `dist/package.json` 
- run `npm publish`

marcjansen:
We should release a new version anyhow, so we cam revisit hiw this is right now.

simonseyock:
The problem was that the package was built to dist and then the current package.json was copied to dist. Then the version bump was performed in the original package.json and the np tried to publish with the old one.

I fixed this with: `"postversion": "cp package.json dist/",`.

PR #580, opened 2020-02-19
Set minimum required node version to >=10

annarieger:
# BREAKING CHANGES
* Introduced `engines` block to `package.json`
* Sets minimum required node version to `>=10`
* Updated `np` package to `v6.1.0` (needs node >10)


LukasLohoff:
Does the release script work again with np 6.1? Currently it is broken and we have to do manual releases (see #503)

annarieger:
I'm not sure about this.. We can try to release the next version to see if it works

PR #582, opened 2020-02-20
Introduce 'activeEpsgCode' for 'CoordinateMousePositionPanel'

chrismayer:
This introduces the config property `activeEpsgCode` for the `BasiGX.view.panel.CoordinateMousePositionPanel` class in order to configure the initially selected EPSG code. Before the last element in the list was set as active EPSG code, which is still the fallback in case `activeEpsgCode` is not provided.


chrismayer:
Thanks for your reviews @annarieger & @dnlkoch ! I'll merge now.

PR #601, opened 2020-05-29
Fix extraction of layers when no bbox is available

hwbllmnn:
Title says it all. @terrestris/devs Please review.


hwbllmnn:
Unsure why the node 10 job fails (doesn't find eslint?!?)

LukasLohoff:
Not sure either, everything looks fine. Maybe the `node_modules` cache?

hwbllmnn:
That was it indeed. Still strange, the `npm i` should have fixed missing packages in the `node_modules` folder...

PR #605, opened 2020-06-10
Change nominatim service to nominatim.terrestris.de

hblitza:
Due to CORS issues with nominatim.openstreetmap.org.

- [x] fix icon issue

### DIFF
--- a/src/store/GazetteerSearch.js
+++ b/src/store/GazetteerSearch.js
@@ -34,7 +34,7 @@ Ext.define('BasiGX.store.GazetteerSearch', {
     _lastRequest: null,
 
     proxy: {
-        url: 'https://nominatim.openstreetmap.org',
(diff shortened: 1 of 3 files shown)


hwbllmnn:
The icon issue is not an issue with the code, https://nominatim.terrestris.de simply does not return icons.

marcjansen:
Can this be merged then? Or is there sth. to be done server side first?

hwbllmnn:
I think we first should see to it that the service returns the icon URLs properly, lest we forget it.

hwbllmnn:
This is ready for review & merge now, turns out the server is running fine but delivering paths instead of URLs.

weskamm:
I like, travis not!

hwbllmnn:
Travis is fixed, merging now

PR #606, opened 2020-06-18
Introduces hoverClick plugin

jansule:
This introduces the hoverClick plugin.

This plugin extends the hover plugin to disable/enable click and hover events separately.

To do so, two configs  `hoverable`, `clickable` were added to enable/disable hovering and clicking globally. These values just set the general options to disable/enable hovering and clicking. A layer  needs to have a truthy `hoverable` or `clickable` property to actually enable hovering and clicking respectively.

Now, separate requests for the click and the hover event will be fired.

Please note that it is required to manually cleanup the feature selection (`hoverClickPlugin.cleanupHoverArtifacts()`) after the `hoverfeaturesclick` event was fired, as in contrast to the `pointerrest` event, we cannot know when the cleanup should be triggered.

### DIFF
--- a/src/plugin/HoverClick.js
+++ b/src/plugin/HoverClick.js
@@ -14,13 +14,13 @@ Ext.define('BasiGX.plugin.HoverClick', {
 
     config: {
         /**
-         * Enable/Disable hovering globally. If true, the plugin will not
+         * Enable/Disable hovering globally. If false, the plugin will not
          * listen to any hover event, regardless if a layer has a truthy
          * hoverable property.
          */
         hoverable: true,
         /**
-         * Enable/Disable clicking globally. If true, the plugin will not
+         * Enable/Disable clicking globally. If false, the plugin will not
          * listen to any click event, regardless if a layer has a truthy
          * clickable property.
          */
@@ -137,7 +137,7 @@ Ext.define('BasiGX.plugin.HoverClick', {
 
             if (source instanceof ol.source.TileWMS
                     || source instanceof ol.source.ImageWMS) {
-                // me.cleanupHoverArtifacts();
+
                 var url = source.getGetFeatureInfoUrl(
                     evt.coordinate,
                     resolution,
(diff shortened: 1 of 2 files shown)


src/plugin/HoverClick.js
```diff
@@ -0,0 +1,262 @@
...
+        /**
+         * The property of a layer that holds a boolean value which indicates
+         * whether this layer qualifies for clicking.
+         */
+        LAYER_CLICKABLE_PROPERTY_NAME: 'clickable'
+    },
+
+    config: {
+        /**
+         * Enable/Disable hovering globally. If true, the plugin will not
```
weskamm, line 17: You are sure with the word `not` in this context?

src/plugin/HoverClick.js
```diff
@@ -0,0 +1,262 @@
...
+         * listen to any hover event, regardless if a layer has a truthy
+         * hoverable property.
+         */
+        hoverable: true,
+        /**
+         * Enable/Disable clicking globally. If true, the plugin will not
```
weskamm, line 23: You are sure with the word `not` in this context?

src/plugin/HoverClick.js
```diff
@@ -0,0 +1,262 @@
...
+
+            var source = layer.getSource();
+            var resolution = mapView.getResolution();
+            var projCode = mapView.getProjection().getCode();
+            var hoverFeaturesRevert = layer.get(hoverFeaturesRevertProp);
+
+
+            if (source instanceof ol.source.TileWMS
+                    || source instanceof ol.source.ImageWMS) {
+                // me.cleanupHoverArtifacts();
```
weskamm, line 140: remove it or fix it :-)

weskamm:
Looking good to me, just added minor comments

src/plugin/HoverClick.js
jansule: Good catch. Solved in https://github.com/terrestris/BasiGX/pull/606/commits/2bfe2df6e23fa75799c175c7ea2187fbfc09436b

src/plugin/HoverClick.js
jansule: Good catch. Solved in https://github.com/terrestris/BasiGX/pull/606/commits/2bfe2df6e23fa75799c175c7ea2187fbfc09436b

src/plugin/HoverClick.js
jansule: solved in https://github.com/terrestris/BasiGX/pull/606/commits/2bfe2df6e23fa75799c175c7ea2187fbfc09436b

jansule:
Thanks for the reviews!

PR #607, opened 2020-06-24
Fix WFS search wrt non standard namespace prefixes

hwbllmnn:
Title says it all. @terrestris/devs Please review.


weskamm:
If it fixes the problem, go for it :-)

hwbllmnn:
> If it fixes the problem, go for it :-)

It does. The namespace prefix that gets passed into the function is actually the one parsed from the describe feature type response, so it should be trusted and preferred :-)

PR #617, opened 2020-09-11
Fix focusable tree item to allow actual actions

hwbllmnn:
This PR adds/removes the `x-focused` class on focusenter/leave and simulates click events when hitting space or enter on focused items.

### DIFF
--- a/src/view/list/FocusableTreeItem.js
+++ b/src/view/list/FocusableTreeItem.js
@@ -73,8 +73,26 @@ Ext.define('BasiGX.view.list.FocusableTreeItem', {
         var pressedKey = evt.getKey();
         var list = me.up('treelist');
         if (list && (pressedKey === enter || pressedKey === space)) {
-            list.fireEvent('selectionchange', list, me.getNode(), {});
+            list.setSelection(me.getNode());
+            try {
+                me.el.dom.click();
+            } catch (e) {
+                Ext.log('Problem when trying to emulate click on node: ' + e);
+            }
         }
+    },
+
+    /**
+     * Binds the focusenter and focusleave events to add/remove the
+     * x-focused class.
+     */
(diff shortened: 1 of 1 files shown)


src/view/list/FocusableTreeItem.js
```diff
@@ -73,10 +73,28 @@ Ext.define('BasiGX.view.list.FocusableTreeItem', {
...
         }
     },
 
+    /**
+     * Binds the focusenter and focusleave events to add/remove the
+     * x-focused class.
+     */
+    bindFocusEvents: function() {
+        this.getEl().on('focusenter', function() {
+            this.dom.classList.add('x-focused');
```
marcjansen, line 91: Untested:

focusenter is fired with the component that gains focus, AFAICT, so this might also work:

```
this.getEl().on('focusenter', function(e, target, fromTarget, toComp, fromComp ) {
  toComp.addCls('x-focused');
})
```
The above does not need the native (and shaky suopported?) `classList`.

Otherwise `this`, or `Ext.get(this)` or `Ext.fly(this)` may be used to get an `Ext.Component`, so that we can then use `addCls`.

Would be very nice if you could give this a try.

marcjansen (approved):
Looking fine, one thing you may wanna try, and depending on the outcome and your decision you can then merge.

src/view/list/FocusableTreeItem.js
marcjansen: See also here: https://caniuse.com/?search=classList IE looks isn't fully green.

hwbllmnn: You're right, I wasn't aware that `classList` isn't fully supported even on IE11. Fixed in 3f776910

annarieger (approved):

PR #618, opened 2020-09-15
Allow use of the layer name instead of feature type for grouping

hwbllmnn:
The WFS search grid groups by feature type name by default. This adds an option to instead use the layer name (determined from `layer.get('name');` instead.

### DIFF
--- a/src/view/grid/MultiSearchWFSSearchGrid.js
+++ b/src/view/grid/MultiSearchWFSSearchGrid.js
@@ -89,6 +89,12 @@ Ext.define('BasiGX.view.grid.MultiSearchWFSSearchGrid', {
         layer: null,
 
         /**
+         * If set to true, the layer name will be used for grouping in the
+         * grid.
+         */
+        useLayerName: false,
+
+        /**
          * Object containing two keys - suffix and prefix - to match begin and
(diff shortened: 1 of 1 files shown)


src/view/grid/MultiSearchWFSSearchGrid.js
```diff
@@ -674,7 +680,11 @@ Ext.define('BasiGX.view.grid.MultiSearchWFSSearchGrid', {
                     }
 
                     feature.properties.displayfield = displayfield;
-                    feature.properties.featuretype = ftName;
+                    if (me.getUseLayerName()) {
+                        feature.properties.featuretype = layer.get('name');
```
annarieger, line 684: Could this be the case, that a certain layer has no name set?

annarieger (approved):

annarieger:
LGTM, just a tiny question regarding possibly not set layer name

src/view/grid/MultiSearchWFSSearchGrid.js
hwbllmnn: Probably, but I'd leave this up to the project using this. If they configured this component to use the layer name they should make sure the name is always set.

PR #621, opened 2020-10-16
Function check out of loop

dockhunter:
This PR suggest to move the function check outside of the loop. 
Please review.


marcjansen (approved):

chrismayer (approved):
Looks good to me. Thank you!

PR #625, opened 2021-01-08
Optional filterFn config for GraphicPool components

annarieger:
Pass optional filter function to `GraphicPool` components which can be used to exclude some records from pool.

Use case:
Pool contains images in different formats, but only PNGs should be displayed.

Additionally adds missed vertical scrollbar to view component.

@terrestris/devs

### DIFF
--- a/src/view/panel/GraphicPool.js
+++ b/src/view/panel/GraphicPool.js
@@ -86,6 +86,12 @@ Ext.define('BasiGX.view.panel.GraphicPool', {
         },
 
         /**
+          * Optional filter function. Can be useful to filter some images in
+          * unsupported format from store.
+        */
+        filterFn: null,
+
+        /**
          * flag that indicates that a csrf-token should be sent to backend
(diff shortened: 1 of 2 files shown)


src/view/panel/GraphicPool.js
```diff
@@ -85,6 +90,12 @@ Ext.define('BasiGX.view.panel.GraphicPool', {
             graphicDelete: null
         },
 
+        /**
+          * Optional filter function. Can be useful to filter some images in
+          * unsupported format from store.
+        */
```
marcjansen, line 96: Asterisks are wrongly indented, it seems.

marcjansen (commented):
Looks good to me. How would the function look like, for your example use case? Maybe a candidate for the docs.

annarieger:
> Looks good to me. How would the function look like, for your example use case? Maybe a candidate for the docs.

Good hint, I've added an example block to the docs.

PR #626, opened 2021-01-13
Multisearch: Show layer on record click

hblitza:
- Sets the layer of a found and clicked record to visible

When a layer is set to unvisible in the clients layer-tree and a search result of that layer is clicked, the map zooms to an unvisible feature on the map. This should fix this behaviour.

Please note @terrestris/devs

### DIFF
--- a/src/view/grid/MultiSearchWFSSearchGrid.js
+++ b/src/view/grid/MultiSearchWFSSearchGrid.js
@@ -775,6 +775,7 @@ Ext.define('BasiGX.view.grid.MultiSearchWFSSearchGrid', {
         var extent;
         var x;
         var y;
+        var layerName = record.getData().featuretype;
 
         layer.getSource().clear();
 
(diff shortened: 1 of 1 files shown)


hwbllmnn (approved):

hblitza:
Thanks for approval.

src/view/grid/MultiSearchWFSSearchGrid.js
```diff
@@ -803,5 +804,12 @@ Ext.define('BasiGX.view.grid.MultiSearchWFSSearchGrid', {
                 olView.setCenter([x, y]);
             }
         }
+        // set layer visibility of clicked feature to true
+        if (layerName) {
+            var olLayer = BasiGX.util.Layer.getLayerByName(layerName);
+            if (!olLayer.getVisible()) {
```
marcjansen, line 810: Can `olLayer` ever become `falsy`, e.g. `undefined`?

hblitza: I am not sure if it can be ever `undefined`. We get the `layerName` from the (successful) response of the WFS-Search, that is based for the respective layer. So it must be within the application. Nevertheless, I will add a check - thanks for your catch!

PR #637, opened 2021-03-12
Deploy gh-pages through github actions

hwbllmnn:
Also deploys versioned gh-pages when tags are pushed (similar to GeoExt).


marcjansen:
Love this, same comments from https://github.com/geoext/geoext3/pull/674 apply here, I guess.

marcjansen:
When you have some spare minutes, it'd be cool if you could compare the actions here with those over at GeoExt, @hwbllmnn. Or is this ready for prime time already?

hwbllmnn:
I think I moved all the changes we made over to this repository, and I don't think we need a build here for the pages, right? Unfortunately the tests seem to be failing, I'm not sure why (and I don't have the time for this right now). I'll have a look when I find the time if noone else volunteers.

marcjansen:
> I think I moved all the changes we made over to this repository, and I don't think we need a build here for the pages, right? Unfortunately the tests seem to be failing, I'm not sure why (and I don't have the time for this right now). I'll have a look when I find the time if noone else volunteers.

I think it failed due to missing `https://geoext.github.io/geoext3/master/GeoExt.js` which should now be there again.

PR #643, opened 2021-05-27
Update ol

simonseyock:
This is a branch to update openlayers. The tests currently fail because geoext is not yet published with the newer openlayers version.

Locally the test fails with one curios error that likely has to do with karma or puppeteer. I will continue to investigate this.


Unraveler (approved):
Looks good to me!

PR #651, opened 2021-06-11
added new methods do do animation on a layer instead of feature. Fixe…

Unraveler:
Adds new methods to Animation component:
- `doAnimation(feature, animate)`
- 'endAnimation(feature, listenerKey)`

They do an animation on a layer instead of a feature, which was causing error when doing the animation

Also fixes Measure.js, where coordinates were incorrectrly passed to Polygon, whereas geometry should be given instead.


weskamm:
Changes looking good to me, but the examples are broken now. Please fix them.

buehner:
yes. changes also look good to me. i could already test them in a project. as @weskamm mentioned, the tests seem to be broken

weskamm:
> yes. changes look good to me, too. i could successfully test them in a project. as @weskamm mentioned, the tests seem to be broken

The fancy examples are broken, not the tests :-)

Unraveler:
Animation and i18n examples are fixed. The update to oOL6.5.0 was missing on the HTML. You can check again @weskamm , @buehner

weskamm:
Looking great again, thanks!

PR #653, opened 2021-07-03
Update measure area tool for ol6

geographika:
From https://openlayers.org/en/latest/apidoc/module-ol_sphere.html - `getArea` expects a geometry rather than an array of coordinates. 
Otherwise the following error is thrown: `e.getType is not a function`


marcjansen:
Code looks nice. Any chance you can add a test that passes now and failed before your change?

marcjansen:
Once the linting is fine this is good to go.

marcjansen (approved):

marcjansen:
Thanks!

PR #654, opened 2021-08-05
Enable button in the addWms form when the textbox is hidden

geographika:
This pull request ensures that the "Request available layers" button becomes enabled if a user enters or pastes a URL in the WMS-URL box. 

Currently if the combobox has a list of entries then the texfField is hidden, but it does not allow blank values, so the form is never valid and the button is disabled (a user however could select something from the dropdown and then paste text as a workaround). 

image


weskamm:
Looks good to me, thanks!

PR #657, opened 2021-08-13
Update dependencies

hwbllmnn:

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,29 +59,29 @@
     },
     "peerDependencies": {
         "@geoext/geoext": "4.0.0",
-        "@geoext/openlayers-legacy": "6.5.2"
+        "@geoext/openlayers-legacy": "6.5.4"
     },
     "devDependencies": {
         "@geoext/geoext": "4.0.0",
-        "@geoext/openlayers-legacy": "6.5.2",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -59,29 +59,29 @@
...
-        "karma-chrome-launcher": "^3.1.0",
+        "karma": "6.3.4",
+        "karma-chrome-launcher": "3.1.0",
         "karma-coverage": "2.0.3",
         "karma-expect": "1.1.3",
         "karma-mocha": "2.0.1",
         "karma-mocha-reporter": "2.2.5",
         "karma-sinon": "1.0.5",
-        "mocha": "8.2.1",
-        "np": "^7.2.0",
```
annarieger, line 79: Any reason why we switched from caret to fixed version here and by some karma-* packages above?

Just a question, in no way a reproach :)

hwbllmnn: Yes, I like to control which versions are actually used, else `npm i` will yield different results at different times. Also, most other packages use fixed versions, so in a way this consolidates it. No reproach received, I actually appreciate criticism, it is oftentimes very helpful :-)

LukasLohoff (approved):

PR #658, opened 2021-08-17
Add optional AJAX proxy URL config for CswForm

chrismayer:
This adds the possibility to configure the `BasiGX.view.form.CSW` component with an optional AJAX proxy URL, which is placed before the "real" target CSW URL.


annarieger (approved):
Nice addition 🍍

weskamm (approved):

chrismayer:
Thanks for your review! I'll merge now.

PR #678, opened 2022-02-15
Change save/edit button state when table or geometry was edited

jansule:
Updates the button state of cancel and save button, if data was edited.

Also had to remove `changefeature` listener, as this will also be called when the style of a feature changes. This always happens when moving/modifying a geometry. Therefore, now the `featurechanged` events of the modify/move buttons also pass through the openlayers event.


Unraveler:
looks good to me

PR #679, opened 2022-02-17
Remove usage of lockFeature on FeatureGrid

jansule:
This removes the usage of the lockFeature request for WFS-T requests, as this poses quite a few problems when used in GeoServer.

This also make better usage of the viewModel.


Unraveler (approved):
LGTM

PR #692, opened 2022-08-03
Also support ImageWMS when filtering

buehner:
The previous code only supported TileWMS as layer source type, but did not support ImageWMS.

This will fix it.

### DIFF
--- a/src/util/WFS.js
+++ b/src/util/WFS.js
@@ -627,6 +627,17 @@ Ext.define('BasiGX.util.WFS', {
             var mapComponent = BasiGX.util.Map.getMapComponent();
             var srs = mapComponent.map.getView().getProjection().getCode();
 
+            var layerSource = filterLayer.getSource();
+            var url;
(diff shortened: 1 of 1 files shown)


hwbllmnn (approved):

src/util/WFS.js
```diff
@@ -627,6 +627,17 @@ Ext.define('BasiGX.util.WFS', {
             var mapComponent = BasiGX.util.Map.getMapComponent();
             var srs = mapComponent.map.getView().getProjection().getCode();
 
+            var layerSource = filterLayer.getSource();
+            var url;
+
+            if(layerSource instanceof ol.source.TileWMS) {
```
LukasLohoff, line 633:
```suggestion
            if (layerSource instanceof ol.source.TileWMS) {
```

LukasLohoff (approved):
LGTM, just a minor linting suggestion.

PR #696, opened 2022-08-09
Introduce support of arcGIS REST FeatureServer layers

jansule:
This adds support for ArcGIS REST FeatureServer layers.

Currently, only importing single layers of a FeatureServer are supported. We get the layer information by firing http requests to all services in an arcGIS folder that are of type FeatureServer. We then take the layer definitions from the responses and add them to the list of checkboxes in the UI. When importing, all features of a single layer will be requested and added as unstyled ol.layer.Vector.

compass-arcgis-rest-featureserver

### DIFF
--- a/src/util/ArcGISRest.js
+++ b/src/util/ArcGISRest.js
@@ -126,7 +126,7 @@ Ext.define('BasiGX.util.ArcGISRest', {
          * @param {string} layerConfig.service.type The service type.
          * @param {string} layerConfig.service.name The service name.
          * @param {string} layerConfig.url The service url.
-         * @param {object} layerConfig.layer (optional) The layer confings of
+         * @param {object} layerConfig.layer (optional) The layer config of
          * a FeatureServer layer.
          * @param {number} layerConfig.layer.id The id of a FeatureServer layer.
          * @param {string} layerConfig.layer.name The name of a FeatureServer
--- a/src/util/ArcGISRest.js
+++ b/src/util/ArcGISRest.js
@@ -94,9 +94,12 @@ Ext.define('BasiGX.util.ArcGISRest', {
          * @param {string} serviceUrl The URL of the service.
          * @param {number} layerId The id of the layer in the FeatureServer.
          * @param {string} format The output format.
+         * @param {string} filter The filter condition.
          * @return {string} The query URL.
          */
-        createFeatureServerQueryUrl: function(serviceUrl, layerId, format) {
+        createFeatureServerQueryUrl: function(
+            serviceUrl, layerId, format, filter
+        ) {
             if (!BasiGX.util.ArcGISRest.isArcGISRestUrl(serviceUrl)) {
                 return;
             }
@@ -112,8 +115,13 @@ Ext.define('BasiGX.util.ArcGISRest', {
             }
             // The query endpoint has a required 'where' parameter. In
             // order to get all features, we apply a where-clause that
-            // always returns true.
-            url = BasiGX.util.Url.setQueryParam(url, 'where', '1=1');
+            // always returns true. This behavior can be overwritten
+            // by specifying a 'filter' argument.
+            var filterToUse = '1=1';
+            if (filter) {
+                filterToUse = filter;
+            }
+            url = BasiGX.util.Url.setQueryParam(url, 'where', filterToUse);
 
             return url;
         },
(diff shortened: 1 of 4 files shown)


src/util/ArcGISRest.js
```diff
@@ -56,6 +61,63 @@ Ext.define('BasiGX.util.ArcGISRest', {
...
+         * @param {string} serviceUrl The URL of the service.
+         * @param {string} serverName The name of the FeatureServer.
+         * @param {string} format The output format.
+         * @return {string} The URL to the FeatureServer.
+         */
+        createFeatureServerUrl: function(serviceUrl, serverName, format) {
+            if (!BasiGX.util.ArcGISRest.isArcGISRestUrl(serviceUrl)) {
+                return;
+            }
+            var urlObj = new URL(serviceUrl);
```
marcjansen, line 76: Not sure about browser compatibility here, `URL` needs a polyfill in ancient browsers (e.g. IE). Is it used elsewhere?

src/util/ArcGISRest.js
```diff
@@ -56,6 +61,63 @@ Ext.define('BasiGX.util.ArcGISRest', {
...
+            var path = parts.join('/');
+
+            var url = urlObj.origin + path;
+            if (format) {
+                url = BasiGX.util.Url.setQueryParam(url, 'f', format);
+            }
+            // The query endpoint has a required 'where' parameter. In
+            // order to get all features, we apply a where-clause that
+            // always returns true.
+            url = BasiGX.util.Url.setQueryParam(url, 'where', '1=1');
```
marcjansen, line 116: Might be a nice addition to have an optional `where`-param, defaulting to `1=1`, this way even filtering becomes possible. Just an idea.

src/util/ArcGISRest.js
```diff
@@ -64,14 +126,20 @@ Ext.define('BasiGX.util.ArcGISRest', {
          * @param {string} layerConfig.service.type The service type.
          * @param {string} layerConfig.service.name The service name.
          * @param {string} layerConfig.url The service url.
+         * @param {object} layerConfig.layer (optional) The layer confings of
```
marcjansen, line 129:
```suggestion
         * @param {object} layerConfig.layer (optional) The layer config of
```
?

src/util/ArcGISRest.js
```diff
@@ -104,10 +173,21 @@ Ext.define('BasiGX.util.ArcGISRest', {
...
                         break;
                     case 'FeatureServer':
-                        // FeatureServers can only be requested
-                        // for single layers in a service.
-                        Ext.log.info('ArcGISRest FeatureServer is ' +
-                            'currently not supported.');
+                        var esrijsonFormat = new ol.format.EsriJSON();
+
+                        var sourceUrl = staticMe.createFeatureServerQueryUrl(
+                            url, layerConfig.layer.id, 'json');
```
marcjansen, line 179: According to the API docs from above, this is optional, and should be checked for existance: `layerConfig.layer` right?

src/view/form/AddArcGISRest.js
```diff
@@ -390,6 +397,92 @@ Ext.define('BasiGX.view.form.AddArcGISRest', {
...
 
+    /**
+     * Loads the layerConfigs for all layers of all FeatureServers.
+     *
+     * @param {object[]} featureServers List of layerConfigs of FeatureServers.
+     * @return {Ext.Promise} A promise that resolves with a list of
+     * layerConfigs, where each layerConfig represents a single layer of a
+     * FeatureServer.
+     */
+    loadFeatureServersLayers: function(featureServers) {
```
marcjansen, line 408: To me this seems hard to read, but this might be more my issue :stuck_out_tongue_winking_eye:

marcjansen (approved):
Thx, only nit-picks, I guess.

src/util/ArcGISRest.js
weskamm: we had a talk about that and as this is not breaking any current behaviour (new feature) we decided to let IE finally go.

geographika: I've also tested in IE using the URL polyfill @jansule added in https://github.com/compassinformatics/cpsi-mapview/pull/582 - and all worked fine.

src/util/ArcGISRest.js
jansule: This is optional for the other switch cases. For this one it is mandatory. I will update the docs accordingly.

src/util/ArcGISRest.js
jansule: Done in https://github.com/terrestris/BasiGX/pull/696/commits/86b61e00656536982ad13a0743063e1719d65032

src/view/form/AddArcGISRest.js
jansule: Done in https://github.com/terrestris/BasiGX/pull/696/commits/86b61e00656536982ad13a0743063e1719d65032. Hope it's better now :grinning:

src/util/ArcGISRest.js
marcjansen: We might simply add a line to the readme, explaining which features we use... and how to get a polyfill for them

marcjansen (approved):

jansule:
Optimized requesting the FeatureServers. @marcjansen it would be great, if you could take another look at it.

hwbllmnn (approved):

PR #698, opened 2022-08-10
Dependency updates and other chore

marcjansen:
This PR suggests to update several dependencies to their latest version (including major upgrades) and also requires more recent `node` and `npm`.

`npm test` (which also lints) is green on my machine.

Please reveiew carefully.


package.json
```diff
@@ -17,8 +17,8 @@
         "ux"
     ],
     "engines": {
-        "node": ">=10 <=12",
-        "npm": ">=6.8.0"
+        "node": ">=16.13",
```
marcjansen, line 20: This is the Gallium LTS

jansule (approved):

PR #699, opened 2022-08-11
Add option to add marker to coordinatemouseposition

jansule:
This adds an optional marker to the map, when the `centerToCoordinate` button was clicked. The marker will be removed with the next map click. By default, the marker uses the default ol style. A custom style can be added via config. Same goes for the layer, where the marker should be added to.

Example with styled marker:
compass-marker-on-coordinate

### DIFF
--- a/src/view/panel/CoordinateMousePositionPanel.js
+++ b/src/view/panel/CoordinateMousePositionPanel.js
@@ -84,6 +84,22 @@ Ext.define('BasiGX.view.panel.CoordinateMousePositionPanel', {
     olMap: null,
 
     /**
+     * The layer for the marker
+     */
+    markerLayer: null,
+
+    /**
+     * True, if marker should be shown at coordinate.
+     * False otherwise.
+     */
+    showMarker: false,
+
+    /**
+     * Optional style for the marker.
+     */
+    markerStyle: null,
+
+    /**
      * The OpenLayers MousePosition control
      */
     olMousePositionControl: null,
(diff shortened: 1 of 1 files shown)


hwbllmnn (approved):

marcjansen:
Have you considered sinply throwing a new event and move all the other logic to the app code?

src/view/panel/CoordinateMousePositionPanel.js
```diff
@@ -455,7 +471,39 @@ Ext.define('BasiGX.view.panel.CoordinateMousePositionPanel', {
...
+     */
+    showMarkerOnMap: function(position) {
+        var me = this;
+        if (!me.markerLayer) {
+            var source = new ol.source.Vector();
+            var style = undefined;
+            if (me.markerStyle) {
+                style = me.markerStyle;
+            }
+            me.markerLayer = new ol.layer.Vector({
```
marcjansen, line 493: Will this layer not show up in legends or layer trees?

jansule: No, in my case it does not show in the layer tree (even when giving a name)

jansule:
> Have you considered sinply throwing a new event and move all the other logic to the app code?

Which logic are you referring to, exactly? The creation of the layer/marker?

marcjansen:
> > Have you considered sinply throwing a new event and move all the other logic to the app code?
> 
> Which logic are you referring to, exactly? The creation of the layer/marker?

Yes. This could all be done in the app. but would be  amajor change of course. This is not totally important.

jansule:
I added an additional check for the layer tree in https://github.com/terrestris/BasiGX/pull/700.

Regarding the layer handling: I would rather keep it as is. It is possible to use a custom layer through the `markerLayer` config anyways. With the current logic, we reduce possible repetitive work at least a little.

PR #703, opened 2022-08-16
Check for empty arrays prior to reduce call in BasiGX.view.form.AddArcGISRest

geographika:
If a server has no FeaureServers then an error is thrown when calling ` Ext.Array.reduce` on an empty array: `TypeError: Reduce of empty array with no initial value`.
This check simply avoids calling the function when `responses` is empty.

### DIFF
--- a/src/view/form/AddArcGISRest.js
+++ b/src/view/form/AddArcGISRest.js
@@ -422,7 +422,7 @@ Ext.define('BasiGX.view.form.AddArcGISRest', {
         return Ext.Promise.all(mappedPromises)
             .then(function(responses) {
                 var configs = [];
-                if (Ext.isEmpty(responses) == false) {
+                if (!Ext.isEmpty(responses)) {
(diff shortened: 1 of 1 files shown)


src/view/form/AddArcGISRest.js
```diff
@@ -421,12 +421,15 @@ Ext.define('BasiGX.view.form.AddArcGISRest', {
...
         return Ext.Promise.all(mappedPromises)
             .then(function(responses) {
-                var configs = Ext.Array.reduce(responses, function(acc, conf) {
-                    if (!conf) {
-                        return acc;
-                    }
-                    return Ext.Array.merge(acc, conf);
-                });
+                var configs = [];
+                if (Ext.isEmpty(responses) == false) {
```
buehner, line 425:
```suggestion
                if (!Ext.isEmpty(responses)) {
```

buehner (changes requested):

buehner (approved):

PR #720, opened 2023-01-30
Allow pre-loading projection definitions to avoid requests to epsg.io

fergaldoyle:
In relation to https://github.com/compassinformatics/cpsi-mapview/issues/615

This PR allows consumers of BasiGX to pre-load epsg.io responses to avoid requests to that service.

It uses a singleton store, `BasiGX.store.Projections` to which definitions can be added. When `fetchProj4jCrsDefinitions` is called it will check if the definition exists in the store, if it exists it will use it, if not it will request the definition from epsg.io.

Definitions can be pre-loaded like this:
```javascript
projectionsStore.loadData([{
    id: 'EPSG:4326',
    code: '4326',
    name: 'WGS 84',
    proj4: '+proj=longlat +datum=WGS84 +no_defs +type=crs',
    unit: 'degree (supplier to define representation)',
}, {
    id: 'EPSG:29902',
    code: '29902',
    name: 'TM65 / Irish Grid',
    proj4: '+proj=tmerc +lat_0=53.5 +lon_0=-8 +k=1.000035 +x_0=200000 +y_0=250000 +a=6377340.189 +rf=299.3249646 +towgs84=482.5,-130.6,564.6,-1.042,-0.214,-0.631,8.15 +units=m +no_defs +type=crs',
    unit: 'metre',
}, {
    id: 'EPSG:2157',
    code: '2157',
    name: 'IRENET95 / Irish Transverse Mercator',
    proj4: '+proj=tmerc +lat_0=53.5 +lon_0=-8 +k=0.99982 +x_0=600000 +y_0=750000 +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs +type=crs',
    unit: 'metre'
}, {
    id: 'EPSG:3857',
    code: '3857',
    name: 'WGS 84 / Pseudo-Mercator',
    proj4: '+proj=merc +a=6378137 +b=6378137 +lat_ts=0 +lon_0=0 +x_0=0 +y_0=0 +k=1 +units=m +nadgrids=@null +wktext +no_defs +type=crs',
    unit: 'metre'
}]);
```

### DIFF
--- a/src/model/Projection.js
+++ b/src/model/Projection.js
@@ -1,0 +1,61 @@
+/* Copyright (c) 2016-present terrestris GmbH & Co. KG
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+/**
+ * Projection mode for use with the Projections Store
+ *
+ * @class BasiGX.model.Projection
+ */
+Ext.define('BasiGX.model.Projection', {
+    extend: 'Ext.data.Model',
+    idProperty: 'code',
+    fields: [{
+        name: 'code',
+        type: 'int'
+    }, {
+        name: 'accuracy',
+        type: 'string'
+    }, {
+        name: 'area',
+        type: 'string'
+    }, {
+        name: 'authority',
+        type: 'string'
+    }, {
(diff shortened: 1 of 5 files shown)


geographika:
We can add some basic tests for the new store to this pull request, but it would be good to get some early feedback on the approach used and it this change will be merged. 

The reasoning behind this change is:

- to be less reliant on the https://epsg.io site - this could be offline, or inaccessible in some environments
- reduce the need to make requests every time the application is loaded to get projection strings when an application may only be using the same 3/4 projections

marcjansen (commented):
I like the idea, and would say continue in the direction you outlined. It would be good if fetched projections would be added to the store, I think. For all practical uses the numerical part of the code is a fine id, so I'd keep this. Obviously some tests would be great.

geographika:
Thanks @marcjansen for the feedback. 
Noted with regards to the code. It looks like even for non-EPSG authorities the codes remain unique e.g. the "ESRI:53016" URL is https://epsg.io/53018 (so 53018 must be unique on the web site).

fergaldoyle:
@marcjansen @geographika 

Updated to use `code` as the identifier, fetched projections are now added to the store, and some tests added.

marcjansen (approved):
Nice this is good to go. Thx.

PR #754, opened 2025-01-17
Check for proj codes as str or int in CoordinateMousePositionPanel

geographika:
The comment states `// Loose equality check since code could be a string or an int` however an exact check is used. 
This then causes `filtered` to be an empty array when checking `3857 ==== "3857" `between a projection code in the store and the projection code from the map. 

Results in the JS error:

```
CoordinateMousePositionPanel.js?_dc=1737122544728:347 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'name')
    at constructor.generateCrsChangeButtonGroup (CoordinateMousePositionPanel.js?_dc=1737122544728:347:53)
    at CoordinateMousePositionPanel.js?_dc=1737122544728:273:20
```

In addition the `filtered ? filtered[0].name : ''` will return `true` for an empty array (which is always created using Ext.Array.filter)  - updated to account for this.

### DIFF
--- a/src/view/panel/CoordinateMousePositionPanel.js
+++ b/src/view/panel/CoordinateMousePositionPanel.js
@@ -339,8 +339,8 @@ Ext.define('BasiGX.view.panel.CoordinateMousePositionPanel', {
                 var mapCode = me.olMap.getView().getProjection().getCode()
                     .split(':')[1];
                 var filtered = Ext.Array.filter(proj4jObjects, function(obj) {
-                    // Loose equality check since code could be a string or an int
-                    return obj.code == mapCode;
+                    // convert values to strings as they can be either a string or an int
+                    return obj.code.toString() == mapCode.toString();
                 });
                 me.getViewModel().setData({
                     srsName: !Ext.isEmpty(filtered) ? filtered[0].name : ''
--- a/src/view/panel/CoordinateMousePositionPanel.js
+++ b/src/view/panel/CoordinateMousePositionPanel.js
@@ -343,7 +343,7 @@ Ext.define('BasiGX.view.panel.CoordinateMousePositionPanel', {
                     return obj.code == mapCode;
                 });
                 me.getViewModel().setData({
-                    srsName: filtered ? filtered[0].name : ''
+                    srsName: !Ext.isEmpty(filtered) ? filtered[0].name : ''
                 });
             }
         }
--- a/src/view/panel/CoordinateMousePositionPanel.js
+++ b/src/view/panel/CoordinateMousePositionPanel.js
@@ -340,7 +340,7 @@ Ext.define('BasiGX.view.panel.CoordinateMousePositionPanel', {
                     .split(':')[1];
                 var filtered = Ext.Array.filter(proj4jObjects, function(obj) {
                     // Loose equality check since code could be a string or an int
-                    return obj.code === mapCode;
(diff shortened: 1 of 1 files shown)


geographika:
@KaiVolland - looking at the history this may be intentional? https://github.com/terrestris/BasiGX/commit/5116f3c7295677a070a678b40a16f442b0bcdf05
However even adding a code as a string, this gets converted to a number as the model `BasiGX.model.Projection` has `code` set to `int`. 

```
        projectionsStore.loadData([{
            code: 4326,
            name: 'WGS 84',
            proj4: '+proj=longlat +datum=WGS84 +no_defs +type=crs',
            unit: 'degree (supplier to define representation)',
        }, {
            code: '3857',
            name: 'WGS 84 / Pseudo-Mercator',
            proj4: '+proj=merc +a=6378137 +b=6378137 +lat_ts=0 +lon_0=0 +x_0=0 +y_0=0 +k=1 +units=m +nadgrids=@null +wktext +no_defs +type=crs',
            unit: 'metre'
        },
```

geographika:
Alternately, the mapCode could be converted to an int:

```
                // update view model to set correct CRS name of map
                var mapCode = parseInt(me.olMap.getView().getProjection().getCode()
                    .split(':')[1], 10);
```

src/view/panel/CoordinateMousePositionPanel.js
```diff
@@ -340,10 +340,10 @@ Ext.define('BasiGX.view.panel.CoordinateMousePositionPanel', {
                     .split(':')[1];
                 var filtered = Ext.Array.filter(proj4jObjects, function(obj) {
                     // Loose equality check since code could be a string or an int
-                    return obj.code === mapCode;
+                    return obj.code == mapCode;
```
KaiVolland, line 343: Wouldn't it be better to use `.toString()` or `Number()` on both?

KaiVolland (approved):

KaiVolland:
@dnlkoch can you please have an additional check and then merge it if it is fine for you.

src/view/panel/CoordinateMousePositionPanel.js
```diff
@@ -339,11 +339,11 @@ Ext.define('BasiGX.view.panel.CoordinateMousePositionPanel', {
                 var mapCode = me.olMap.getView().getProjection().getCode()
                     .split(':')[1];
                 var filtered = Ext.Array.filter(proj4jObjects, function(obj) {
-                    // Loose equality check since code could be a string or an int
-                    return obj.code === mapCode;
+                    // convert values to strings as they can be either a string or an int
+                    return obj.code.toString() == mapCode.toString();
```
dnlkoch, line 343: If both values are being converted to a string, I would prefer a strict equality check:

```suggestion
                    return obj.code.toString() === mapCode.toString();
```

And just wondering if `obj.code` is defined always? Otherwise `toString()` might fail.

dnlkoch: Yes, good idea! To also match `undefined` we might want to use `Ext.isDefined` (or whatever is appropriate), something like:

```js
Ext.isDefined(obj.code) && (obj.code.toString() === mapCode.toString())
```

dnlkoch (approved):

dnlkoch:
Thanks @geographika!

